PR #43, opened 2026-03-14
feat: parallelize safe read-only tool calls

furukama:
## Summary
- add a container-side parallel tool execution policy for safe read-only tool calls
- batch eligible tool calls within a single turn while preserving tool result order in history
- add focused tests for parallel eligibility and ordered concurrent execution

## Testing
- npx biome check --write container/src/index.ts container/src/tool-parallelism.ts tests/container.tool-parallelism.test.ts
- npm --prefix container run lint
- npm run lint
- ./node_modules/.bin/vitest run --config vitest.unit.config.ts tests/container.tool-parallelism.test.ts
- npm run build


furukama:

PR #49, opened 2026-03-14
feat: notify HybridAI when a bot is used by HybridClaw

snoller:
On bot set, PATCH /api/bots/settings/<bot_id> with usedfor=HybridClaw:<agentId>. On bot change or session reset, clear the previous bot's usedfor flag. All API calls are fire-and-forget to avoid blocking the user flow.


furukama (commented):
Glaube, das brauchen wir nicht, da wir ja jeden Turn sowieso an HAI schicken.

PR #84, opened 2026-03-18
feat: add in-loop context compaction guard

furukama:
## Summary
- add a container-side three-tier context guard before every model call
- add in-loop compaction with compression-model summarization and heuristic fallback
- pass guard settings through runtime config and container IPC, and cover the new paths with focused tests

## Testing
- npm run test:unit -- tests/context-guard.test.ts tests/in-loop-compaction.test.ts tests/token-usage.cache.test.ts
- npm run lint
- npm --prefix container run lint
- npm run build

## Notes
- gateway-side pre-turn session compaction and pre-compaction memory flush remain unchanged; this PR adds the in-loop safety net for long autonomous turns


furukama:

PR #85, opened 2026-03-18
Fix browser click fallbacks for JS-only cards

furukama:
## Summary
- add selector and visible-text fallback support to `browser_click` when snapshot refs are missing
- keep existing ref-based clicks unchanged while returning target metadata for fallback clicks
- add focused tests covering ref, selector, and visible-text click modes

## Validation
- npm run test:unit -- tests/browser-snapshot-command-args.test.ts tests/browser-click-command-args.test.ts
- npm run build


furukama:

PR #138, opened 2026-03-26
feat: add jobs board and improve message delivery

furukama:
## Summary
- add the admin jobs kanban board, inline job editing, drag and drop ordering, and related scheduler/runtime support
- improve the shared message tool so email sends support subject, cc, and bcc while keeping attachment delivery on the same path
- render markdown tables in the static web chat and add coverage for the email message payload flow

## Validation
- ./node_modules/.bin/vitest run tests/message.tool-actions-send.test.ts tests/email.channel.test.ts
- npm run typecheck
- npm run build

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
 # Changelog
 
-## [Unreleased]
+## [Coming up]
+
+### Changed
+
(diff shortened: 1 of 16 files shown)


furukama:

PR #144, opened 2026-03-27
[codex] Clarify runtime model metadata

furukama:
## Summary
Clarify the runtime metadata injected into the system prompt so model identity answers mention the served model and backend more naturally.

## What Changed
- replace the old runtime metadata model line with a `Model: ... served through ...` sentence
- strip serving-provider prefixes from direct provider model ids where helpful
- preserve upstream vendor labels for routed model ids such as `openrouter/deepseek/deepseek-v3.2`
- add prompt-hook coverage for both direct and routed model formatting

## Why
The previous prompt metadata exposed the raw model id, which made answers to simple questions like "what model are you running?" less readable than necessary.

## Validation
- `./node_modules/.bin/vitest run tests/prompt-hooks.tool-summary.test.ts`


furukama:

PR #229, opened 2026-04-05
[codex] Add Anthropic provider with API key and Claude CLI modes

furukama:
## What changed

This adds Anthropic as a first-class HybridClaw provider with two login methods:

- `--method api-key` for direct Anthropic Messages API access
- `--method claude-cli` for the official `claude -p` transport

The Claude CLI path was simplified to the common-case design:

- it is host-sandbox only
- it shells out to the locally installed `claude` binary
- it reuses the user's existing `claude auth login` session
- container sandbox mode fails fast with a clear error instead of trying to synthesize Claude credentials inside Docker

The PR also fixes a few real Claude CLI compatibility issues discovered during manual smoke testing:

- current Claude login detection on macOS keychain
- no unsupported `--cwd` flag
- `--verbose` added for `--output-format stream-json`

## Why

Direct reuse of Claude Code login material inside a third-party Anthropic API transport is no longer a good assumption. The simpler and more correct split is:

- `api-key` for direct API access
- `claude-cli` for official Claude Code execution semantics

Restricting `claude-cli` to host sandbox mode keeps the implementation smaller and aligned with the main user path of an already-installed Claude CLI on the host.

## Impact

Users can now configure Anthropic with either:

- `hybridclaw auth login anthropic --method api-key --set-default`
- `hybridclaw auth login anthropic --method claude-cli --set-default`

For `claude-cli`, the intended runtime is:

- `hybridclaw gateway start --foreground --sandbox=host`

Container sandbox mode remains supported for Anthropic API-key usage, but not for the Claude CLI transport.

## Validation

- `npm run format`
- `npm run typecheck`
- `npm run build`
- `npm run test:unit -- tests/container.anthropic-claude-cli.test.ts tests/providers.task-routing.test.ts tests/cli.test.ts`

## Remaining risk

This was not fully live-tested end-to-end against every Claude CLI version. The remaining risk is CLI contract drift across installed Claude versions, although the current implementation was adjusted against a real local Claude install during smoke testing.


furukama:

PR #272, opened 2026-04-10
[codex] Normalize provider model prefix stripping

furukama:
## What changed

This change centralizes provider-prefix stripping in a shared `stripProviderPrefix` helper and reuses it across HybridAI, Codex, and Ollama model normalization paths.

It also fixes the host auxiliary HybridAI request path so display-only model names like `hybridai/gpt-5-nano` are normalized to the upstream model id before the API request is sent.

The test coverage now includes the shared helper behavior, the HybridAI auxiliary regression, and OpenAI-compatible Codex request normalization.

## Why

Provider-prefixed display names were being normalized inconsistently across runtime paths. The HybridAI auxiliary path could still send `hybridai/...` to the API, and Codex/Ollama prefix stripping logic was duplicated in multiple places.

## Impact

HybridAI-backed auxiliary text calls now send the runtime model id expected by the API.

Codex and Ollama request normalization now reuse the same shared helper, which reduces drift between providers and makes future prefix handling easier to keep consistent.

## Validation

- `./node_modules/.bin/vitest run --configLoader runner --config vitest.unit.config.ts tests/providers.auxiliary.test.ts tests/model-names.test.ts tests/openai-compatible-model.test.ts`
- `npm run typecheck`

### DIFF
--- a/container/shared/model-names.d.ts
+++ b/container/shared/model-names.d.ts
@@ -5,6 +5,11 @@ export declare function hasKnownNonHybridProviderPrefix(model: string): boolean;
 
 export declare function hasDisplayOnlyHybridAIPrefix(model: string): boolean;
 
+export declare function stripProviderPrefix(
+  model: string,
+  prefix: string,
+): string;
+
 export declare function stripHybridAIModelPrefix(model: string): string;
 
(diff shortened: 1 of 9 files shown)


furukama:

PR #276, opened 2026-04-10
[codex] add per-agent skill filtering

furukama:
## What changed
- added a first-class `skills` field to agent config and registry state
- applied per-agent skill filtering in skill loading with these semantics:
  - omitted or `null`: inherit globally enabled skills
  - `[]`: disable all skills for that agent
  - non-empty list: allow only the listed skills
- persisted agent skill allowlists in SQLite and added schema migration v18
- carried the field through admin agent API handling and `.claw` agent manifest/archive round-tripping
- added targeted coverage for registry persistence, runtime filtering, admin API handling, and archive round-trip behavior

## Why
HybridClaw already supported global and per-channel skill toggles, but there was no way to narrow skills by agent. That made specialized agents inherit the full enabled skill set, which is broader than needed and diverges from the per-agent filtering model used in deer-flow.

## Impact
Agents can now define their own skill allowlists without affecting other agents or channel-level skill configuration. Global and per-channel disabled skills still apply, and the agent-level list only narrows the available set.

## Validation
- `npm run typecheck`
- `npm run lint`
- `./node_modules/.bin/vitest run --configLoader runner --config vitest.unit.config.ts tests/agent-registry.test.ts tests/gateway-http-server.test.ts tests/unit/claw-archive.test.ts`
- `./node_modules/.bin/vitest run --configLoader runner --config vitest.integration.config.ts tests/skill-resolution.integration.test.ts`


furukama:

Issue #249, opened 2026-04-07
One-off jobs


furukama:
Currently, we can create jobs with cron / every / at. But we need a one shot job type. That just gets started once and if successful moves to in review stage. And if not -> retry 3 times (configurable) and then move to in review (failed).

PR #278, opened 2026-04-10
[codex] Add one-shot scheduler jobs

furukama:
## Summary
- add a first-class `one_shot` scheduler job kind with configurable `maxRetries`
- move one-shot jobs into `review` on success or after exhausting retries instead of leaving them stuck in progress or auto-disabled
- expose the new job type and retry setting in the admin scheduler UI and job board

## Why
Issue #249 asks for one-off jobs that run once, retry on failure a limited number of times, and then land in review whether they succeeded or failed. The existing scheduler only approximated that behavior through `at` plus board-state conventions, which left failures in an inconsistent state and did not allow per-job retry limits.

## Impact
- operators can create explicit one-shot jobs from /admin/scheduler
- one-shot jobs retry once per minute until `maxRetries` is exhausted
- moving a run-once job back to backlog resets its runtime state so it can be queued again cleanly
- legacy board-managed `at` jobs keep working and now reconcile to review after terminal failure

## Validation
- `./node_modules/.bin/vitest run tests/scheduler.backlog-retry.test.ts tests/gateway-service.scheduler.test.ts`
- `npm --workspace console run test -- scheduler.test.tsx`
- `npm run typecheck`
- `npm run lint`
- `npm run format`

Refs #249


furukama:

PR #282, opened 2026-04-12
feat(console): replace inline banners with toast notifications

maxnoller:
## Summary

- Replace inline success/error banners with non-blocking toast notifications across all console routes, so feedback no longer pushes page content around or gets lost below the fold
- Toasts auto-dismiss, pause on hover, and stack with overflow management — errors use `role="alert"` for screen reader urgency
- Add reusable Dialog and Toast primitives with full a11y (focus trap, scroll lock, escape-to-close, `aria-live`)

## Test plan

- [ ] Trigger save/delete/pause actions across routes (channels, config, gateway, jobs, mcp, models, scheduler, sessions, skills, email) — toast appears with correct success/error messaging
- [ ] Toast auto-dismisses after ~5s; hovering pauses the timer
- [ ] At most 3 toasts visible — oldest exits when a 4th fires
- [ ] Screen reader announces errors immediately (`role="alert"`) and success/info politely (`role="status"`)
- [ ] Dialog opens/closes via trigger, backdrop click, and Escape
- [ ] Animations respect `prefers-reduced-motion`
- [ ] `vitest` passes (new + existing tests)

### DIFF
--- a/console/src/components/dialog/dialog.test.tsx
+++ b/console/src/components/dialog/dialog.test.tsx
@@ -1,0 +1,101 @@
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { useState } from 'react';
+import { describe, expect, it, vi } from 'vitest';
+import {
+  Dialog,
+  DialogClose,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from './index';
+
+function TestDialog(props: { onConfirm?: () => void }) {
+  const [open, setOpen] = useState(false);
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
+      <DialogTrigger>Open</DialogTrigger>
+      <DialogContent>
+        <DialogHeader>
+          <DialogTitle>Confirm</DialogTitle>
+          <DialogDescription>Are you sure?</DialogDescription>
+        </DialogHeader>
+        <DialogFooter>
+          <DialogClose>Cancel</DialogClose>
+          <button
+            type="button"
+            onClick={() => {
+              props.onConfirm?.();
+              setOpen(false);
+            }}
+          >
+            Yes
+          </button>
+        </DialogFooter>
+      </DialogContent>
+    </Dialog>
+  );
+}
+
+describe('Dialog', () => {
+  it('opens when trigger is clicked and closes on DialogClose', () => {
+    render(<TestDialog />);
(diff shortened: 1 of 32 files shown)


maxnoller:
### Code review

Found 4 issues:

1. `useHideOthers` marks the dialog backdrop `inert`, breaking click-to-close. The dialog portal renders `backdrop` and `viewport` as sibling children of `document.body`. `useHideOthers(panelRef, open)` iterates `document.body.children` and sets `inert = true` on every element that does not contain `panelRef`. The viewport contains `panelRef` (skipped), but the backdrop is a sibling that does not (marked inert). With `inert`, the backdrop becomes non-interactive and the `onClick` handler on line 181 can never fire. The old `aria-hidden` approach did not suppress pointer events.

https://github.com/HybridAIOne/hybridclaw/blob/17d83fa349e6f2d8a2dc9d3bee0bc8cb9e334eea/console/src/hooks/useHideOthers.ts#L22-L27

https://github.com/HybridAIOne/hybridclaw/blob/17d83fa349e6f2d8a2dc9d3bee0bc8cb9e334eea/console/src/components/dialog/index.tsx#L154-L184

2. `preventCloseOnOutsideClick` prop on `DialogContent` has no caller (AGENTS.md 3.2: "Do not add config keys, interfaces, or feature flags without a concrete caller.")

https://github.com/HybridAIOne/hybridclaw/blob/17d83fa349e6f2d8a2dc9d3bee0bc8cb9e334eea/console/src/components/dialog/index.tsx#L130-L132

3. `toast.update()` method and return-value IDs have no caller. `update` is fully implemented, every `add`/`success`/`error`/`info` method returns the toast ID to support it, but no call site in the PR uses `update`, `dismiss`, or captures an ID. (AGENTS.md 3.2: "Do not add config keys, interfaces, or feature flags without a concrete caller.")

https://github.com/HybridAIOne/hybridclaw/blob/17d83fa349e6f2d8a2dc9d3bee0bc8cb9e334eea/console/src/components/toast/index.tsx#L64-L73

4. `size` prop on `DialogContent` supports `'sm' | 'default' | 'lg'` but only `'sm'` is used in the PR. The `'lg'` variant and its CSS class have no caller. (AGENTS.md 3.2: "Do not design for hypothetical future requirements.")

https://github.com/HybridAIOne/hybridclaw/blob/17d83fa349e6f2d8a2dc9d3bee0bc8cb9e334eea/console/src/components/dialog/index.tsx#L126-L128

PR #287, opened 2026-04-13
feat(console): improve empty states for Jobs, Scheduler, Plugins, and MCP

maxnoller:
## Summary

This PR replaces bare, low-information empty states across four admin console pages with helpful, actionable alternatives that include a description of what the page is for and a clear CTA. A pre-existing crash in `channels-catalog.ts` was found and fixed during browser verification.

---

### Jobs (`/admin/jobs`)
- **Before**: Five empty kanban columns each showing "No jobs".
- **After**: When the board has no jobs at all, the columns are replaced by a single centered empty state:
  - _"Jobs are async tasks created by the agent or triggered manually."_
  - **New Job** button (links to `/admin/scheduler`, same as the existing header button).
  - Individual column "No jobs" placeholders are preserved for when there *are* jobs but a column happens to be empty.

### Scheduler (`/admin/scheduler`)
- **Before**: "No scheduled work yet." with the full job editor always visible on the right.
- **After**:
  - Empty list shows _"Scheduled jobs let you run agent tasks automatically on a cron schedule."_ and a **New job** button.
  - The editor panel on the right is hidden until the user clicks **New job** (in the header or empty state) or selects an existing job. Clicking Cancel with no job selected hides the editor again (in place — no navigation).
  - When the editor is hidden, the page collapses to a single-column layout so the jobs list uses the full width.

### Plugins (`/admin/plugins`)
- **Before**: "No plugins match this filter." for both "no plugins installed" and "filter has no results".
- **After**:
  - When the server reports zero installed plugins: centered empty state with _"Plugins extend HybridClaw with custom commands, tools, and hooks."_ and a **Plugin documentation** link (`hybridclaw.io/docs/extensibility/plugins`).
  - When a filter is active and returns no results: existing "No plugins match this filter." message is kept.

### MCP (`/admin/mcp`)
- **Before**: Full server editor form always visible even when no servers are configured.
- **After**:
  - When no servers exist and the editor is not open: full-page centered empty state with _"MCP servers let the agent call external tools over the Model Context Protocol."_ and an **Add your first server** button.
  - Clicking the button reveals the editor. Once at least one server exists the normal list + editor two-column layout is used.
  - The **New server** button in the header is hidden when there are no servers (the empty-state CTA takes its place). It reappears once servers exist.
  - When servers exist but no editor is open, the page collapses to a single-column layout so the list uses the full width.

### CSS additions (`styles.css`)
Two new utility classes following the project's existing conventions:
- `.page-empty` — full-page centered flex layout (used for Jobs board-level and MCP page-level empty states).
- `.empty-state-cta` — panel-level centered bordered block with description + CTA button (used inside list panels for Scheduler and MCP).

### Bugfix: `channels-catalog.ts` crash on missing integration keys
During browser testing, the Scheduler page crashed with `Cannot read properties of undefined (reading 'enabled')` in `describeSlack`. The `describe*` functions assumed every integration key (`slack`, `discord`, `telegram`, etc.) is always present in the config object, but they are absent when the integration has never been configured. Fixed all seven functions with explicit guards / nullish coalescing so they return an "available / not configured" state gracefully.

---

## Test plan

- [x] `/admin/jobs` — no jobs → centered empty state with "New Job" button; clicking goes to `/admin/scheduler`. Verified in browser.
- [x] `/admin/scheduler` — no jobs → list shows CTA; editor panel is hidden. Clicking "New job" (header or CTA) reveals the editor. Verified in browser.
- [x] `/admin/plugins` — no plugins → description + docs link in registry panel. Verified in browser.
- [x] `/admin/mcp` — no servers → full-page centered empty state; "Add your first server" opens the editor. Verified in browser.
- [ ] `/admin/jobs` with some jobs — kanban renders normally; empty columns show the dashed "No jobs" placeholder.
- [ ] `/admin/scheduler` with existing jobs — list renders, clicking a job opens the editor.
- [ ] `/admin/plugins` with plugins but an active filter matching nothing — "No plugins match this filter." shown (not the docs CTA).
- [ ] `/admin/mcp` with servers — normal list + editor layout with "New server" header button.
- [ ] `/admin/channels` — channel catalog renders without crash on instances where some integrations are absent from config.


maxnoller:
### Code review

Found 3 issues:

1. **`describeDiscord`: `enabled` is always `true` when `config.discord` is absent** — the `?? false` fallback is dead code. `config.discord?.groupPolicy !== 'disabled'` evaluates `undefined !== 'disabled'`, which is `true`, so the `||` yields `true` before `??` can fire. When Discord has no config section, the channel shows as `configured` (not `available`) and `active` is gated only on `tokenConfigured`. Fix: guard with `config.discord ? (...) : false`.

https://github.com/HybridAIOne/hybridclaw/blob/4d4dd862e750f946eff7cd1d792b19bc9cee5ffc/console/src/routes/channels-catalog.ts#L62-L70

2. **`describeWhatsApp`: same `!== 'disabled'` flaw with no guard** — when `config.whatsapp` is absent but `options.whatsappLinked === true`, `enabled` evaluates to `true` (both `undefined !== 'disabled'` comparisons are `true`) and `statusTone` becomes `'active'`. The summary line would render `"Linked device · groups undefined"`.

https://github.com/HybridAIOne/hybridclaw/blob/4d4dd862e750f946eff7cd1d792b19bc9cee5ffc/console/src/routes/channels-catalog.ts#L98-L110

3. **`mcp.tsx`: `showEditor` not reset on server delete** — `deleteMutation.onSuccess` calls `setSelectedName(null)` and `setDraft(createDraft())` but never `setShowEditor(false)`. After deleting the last server, `showEditor` remains `true`, the full-page empty state condition (`!showEditor`) is never satisfied, and the UI shows a hollow two-column grid with a blank editor instead of the intended "Add your first server" prompt.

https://github.com/HybridAIOne/hybridclaw/blob/4d4dd862e750f946eff7cd1d792b19bc9cee5ffc/console/src/routes/mcp.tsx#L127-L138

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

maxnoller:
Addressed the review feedback in fe84960a and 40e60d89:

- **`describeTelegram` `inboundEnabled` flaw** — replaced optional chaining with an explicit `config.telegram ? ... : false` guard so an absent telegram config section no longer evaluates `undefined !== 'disabled'` as `true`. (`describeDiscord` and `describeWhatsApp` already had this fix from the prior review pass.)
- **Scheduler two-column grid leaves empty right column** — the outer wrapper now uses `two-column-grid` only when `showEditor` is true, otherwise it falls back to `page-stack` so the jobs list can use the full width.
- **Scheduler Cancel still navigated to `/admin/jobs`** — removed the `window.location.href` redirect in the cancel handler for unsaved new jobs. Cancel now just closes the editor in place.
- **MCP same two-column issue** — applied the same `showEditor ? 'two-column-grid' : 'page-stack'` switch on `/admin/mcp`, so once servers exist the list still gets full width until the user opens the editor.
- **CSS class naming (`.page-empty` vs `.jobs-board-empty`)** — kept the class as `.page-empty` since it is reused by both the Jobs and MCP page-level empty states; updated the PR description to match.

The Copilot comment about `showEditor` defaulting to `false` on `/admin/mcp` is intentional: the empty-state-vs-grid branch hinges on `mcpQuery.data?.servers.length`, not `showEditor`. With servers present, the two-column-grid branch is always taken; the right column is then conditionally hidden until the user picks a server (now via single-column layout).

Tests still pass (`npm --workspace console test` → 311/311).

PR #289, opened 2026-04-13
feat: add confirmation dialogs for destructive admin actions

maxnoller:
## Summary

- **New component**: `console/src/components/ui/alert-dialog.tsx` (+ CSS module) — a self-contained `AlertDialog` built from scratch on the existing custom accessibility hooks (`useFocusTrap`, `useScrollLock`, `useEscapeKeydown`, `useHideOthers`, `useExitAnimation`). Uses `role="alertdialog"` for correct ARIA semantics. Follows shadcn/Base UI compound-component naming with `AlertDialogAction` supporting a `variant="destructive"` prop that maps to the existing `danger-button` CSS class.
- **Restart Gateway** (`/admin/gateway`): click now opens a confirmation dialog before the mutation fires.
- **Delete session** (`/admin/sessions`): existing dialog replaced with `AlertDialog`; copy updated to match spec; confirm button uses `variant="destructive"`.
- **Forget token** (sidebar footer): click now opens a confirmation dialog before `onLogout` is called; confirm button uses `variant="destructive"`.

No new dependencies added.

## Test plan

- [x] Navigate to `/admin/gateway` → click "Restart Gateway" → dialog appears with correct title/description → "Restart" fires the mutation; "Cancel" closes without firing
- [x] Navigate to `/admin/sessions` → select a session → click "Delete session" → dialog shows updated copy and red Delete button → confirm deletes; cancel does not
- [ ] Click "Forget token" in the sidebar footer → dialog appears → confirm logs out; cancel keeps the session
- [ ] Escape key and backdrop click close all three dialogs without triggering the action
- [ ] Tab focus cycles within each dialog; screen reader announces `role="alertdialog"` with correct label/description


maxnoller:
### Code review

Found 1 issue:

1. The Restart Gateway confirm button uses `primary-button` (blue) while the trigger button, the Delete session confirm, and the Forget token confirm all use `danger-button` (red). A gateway restart interrupts all active sessions — it should carry the same visual weight as the other destructive actions in this PR.

https://github.com/HybridAIOne/hybridclaw/blob/48db9087aa8307550fbd4adf49019fdeea5f51b9/console/src/routes/gateway.tsx#L291-L301

PR #292, opened 2026-04-13
[codex] Add TUI approval picker

furukama:
## What changed

This updates the TUI approval flow from a plain numeric prompt to an interactive approval picker.

- adds a focused approval modal with up/down navigation, enter-to-confirm, number-key quick select, and Esc cancel
- reshapes the approval presentation to read more like a Claude-style approval card
- preserves the existing text prompt fallback for non-interactive terminals

## Why

The previous flow worked, but it was slower and less ergonomic than modern terminal coding tools. Approval prompts are a frequent interaction, so the TUI should support direct keyboard navigation instead of requiring typed responses.

## Impact

Users in the local TUI can handle approvals more quickly and with less typing while keeping the same approval policy behavior.

## Validation

- ./node_modules/.bin/vitest run tests/tui-approval-prompt.test.ts tests/tui-approval.test.ts tests/tui-skill-config.test.ts
- npm run typecheck
- npm run lint

## Notes

I did not run a manual npm run tui smoke test in an interactive terminal session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## [Coming up]
 
+### Changed
+
+- **Interactive TUI approval picker**: Local TUI approval requests open a
+  keyboard-driven picker with `Up`/`Down` navigation, `Enter` confirmation,
+  number-key quick select, `Esc` to skip, and a text fallback for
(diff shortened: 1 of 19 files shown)


furukama:

PR #298, opened 2026-04-13
[codex] fix email cursor replay on first sync

furukama:
## What changed
Fix the email IMAP cursor bootstrap logic so a missing or incompatible persisted cursor seeds from the current mailbox head instead of replaying the entire inbox.

Add regression coverage for both startup modes:
- first start with no saved cursor does not ingest old inbox mail
- restart with a saved cursor still processes messages that arrived while HybridClaw was offline

## Why
After the April 11, 2026 email cursor change, the first gateway start without an existing cursor file initialized `lastProcessedUid` to `0` and searched all mailbox UIDs. That caused historical inbox mail to be treated as new and replied to again.

## Impact
Email users no longer get duplicate replies after upgrading into the persisted-cursor implementation.

## Root cause
The persisted cursor path did not preserve the old channel behavior of treating the current mailbox head as the startup baseline when no cursor existed yet.

## Validation
- `./node_modules/.bin/vitest run --configLoader runner --config vitest.unit.config.ts tests/email.connection.test.ts`

### DIFF
--- a/src/channels/email/connection.ts
+++ b/src/channels/email/connection.ts
@@ -246,12 +246,23 @@ export function createEmailConnectionManager(
     try {
       const uidValidity = resolveMailboxUidValidity(activeClient);
       const storedCursor = persistedCursorState.get(folder);
+      const hasStoredCursor =
+        Boolean(storedCursor) && storedCursor.uidValidity === uidValidity;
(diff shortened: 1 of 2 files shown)


furukama:

PR #301, opened 2026-04-13
[codex] add network policy cli

furukama:
## What changed
- added a `hybridclaw policy` CLI for status, list, allow, deny, delete, reset, default, and preset management
- added `/policy` slash-command support plus bundled network policy presets
- replaced the flat `trusted_network_hosts` handling with structured network rules and first-match runtime evaluation
- introduced a shared network-policy normalization/migration module used by both host-side tooling and the container approval runtime
- stored preset provenance on preset-managed rules so `policy preset remove` only removes rules that preset added

## Why
Operators had to hand-edit `.hybridclaw/policy.yaml`, and the original host/container implementation drifted into separate policy cores while the preset removal path could delete manual rules that happened to match a preset.

## Impact
- operators can manage HTTP egress policy from the CLI and slash command instead of editing YAML directly
- runtime enforcement now supports ordered allow/deny rules scoped by host, port, method, path, and agent
- preset application/removal is reversible for preset-owned rules without clobbering identical manual entries

## Root cause
The previous implementation used a flat trusted-host list and duplicated policy normalization/migration logic in separate modules. Preset removal inferred ownership from rule equality rather than explicit provenance.

## Validation
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/approval-policy.test.ts tests/policy-store.test.ts tests/policy-cli.test.ts`


furukama:

PR #307, opened 2026-04-13
chore: replace hand-rolled Docker tag scripts with metadata-action

maxnoller:
## Summary

- Replaces the duplicated bash "Resolve agent/gateway image tags" steps in `publish-release.yml` (`publish-agent` and `publish-gateway` jobs) with `docker/metadata-action@v5.7.0`
- Replaces the "Resolve image tags" bash step in `docker-build.yml` with the same action
- Eliminates ~40 lines of brittle shell that manually constructed tag strings and implemented stable-vs-pre-release `latest`-suppression logic
- All downstream steps (`docker/build-push-action`, `verify-published-tags`) continue to consume `steps.meta.outputs.tags` unchanged — no wiring changes needed

## Behaviour preserved

| Scenario | Before | After |
|---|---|---|
| Stable release (`v1.2.3`) | GHCR `:v1.2.3` + `:latest`; DockerHub same if secrets set | Same via `type=raw,value=latest,enable=true` |
| Pre-release (`v1.2.3-rc.1`) | GHCR `:v1.2.3-rc.1` only | Same via `enable=${{ !contains(..., '-') }}` |
| DockerHub | Added when `publish_dockerhub == 'true'` | Same via conditional image line |
| Main branch builds | `sha-<short>` + `main` | Same via `type=sha,prefix=sha-,format=short` + `type=raw,value=main` |

## Test plan

- [ ] Trigger a `workflow_dispatch` run of Publish Release with a pre-release tag and verify only the versioned tag is pushed (no `latest`)
- [ ] Trigger with a stable tag and verify both versioned and `latest` tags appear on GHCR
- [ ] Verify a main-branch push produces `sha-<short>` and `main` tags via Docker Build workflow

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -33,16 +33,19 @@ jobs:
       - name: Checkout
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
-      - name: Resolve image tags
+      - name: Compute lowercase owner
+        env:
+          REPO_OWNER: ${{ github.repository_owner }}
+        run: echo "OWNER_LC=$(echo "$REPO_OWNER" | tr '[:upper:]' '[:lower:]')" >> "$GITHUB_ENV"
+
+      - name: Docker metadata
         id: meta
-        shell: bash
-        run: |
-          set -euo pipefail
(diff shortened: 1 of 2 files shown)


maxnoller:
Closing — the additional dependency and complexity of metadata-action outweighs the benefit of removing the inline bash. The original tag scripts are self-contained and auditable; the fix for the empty-string conditional can be done as a targeted patch to the bash if needed.

PR #313, opened 2026-04-14
[codex] centralize gateway and skill import errors

furukama:
## What changed
- added a shared `AppError` base and `HttpStatusError` hierarchy under `src/errors/`
- centralized skill import failures in `src/skills/skill-errors.ts` and removed the three duplicated local `SkillImportError` definitions
- unified gateway request/status errors so `gateway-service.ts` and `gateway-http-server.ts` both use the same `GatewayRequestError`
- added a focused hierarchy test to lock in the new inheritance boundaries

## Why
The codebase had several locally defined custom errors with overlapping roles and no shared base. The immediate DRY issues were the triplicated `SkillImportError` classes and the separate `GatewayRequestError` / `HttpRequestError` implementations even though both represented HTTP-status failures.

## Impact
- gateway error handling now relies on one shared status-carrying error type
- skill import paths share a single canonical import error class and not-found subclass
- future error consolidation has a stable base to build on without repeating constructor boilerplate

## Root cause
Error types were being introduced inline within individual modules instead of being treated as shared domain primitives, so similar concerns drifted into separate local classes.

## Validation
- `npm run format -- tests/error-hierarchy.test.ts`
- `npm run typecheck`
- `npm exec vitest -- run --configLoader runner --config vitest.unit.config.ts tests/error-hierarchy.test.ts tests/skills-import.test.ts tests/gateway-service.admin-skills.test.ts tests/gateway-http-server.test.ts`


furukama:

PR #314, opened 2026-04-14
[codex] Add doctor-driven resource hygiene maintenance

furukama:
## What changed
- add doctor-backed resource hygiene checks for stale workspaces, old temp media, old run logs, session compaction backlog, and orphaned exports / prompt dumps
- add a maintenance runner that executes those hygiene checks, auto-applies only safe `warn` fixes, records audit/log summaries, and flags approval-required cleanup separately
- wire a built-in twice-daily `resource-hygiene` scheduler system event into the default runtime config and `config.example.json`
- teach doctor fixes to declare `requiresApproval` so non-interactive `doctor --fix` skips risky remediations instead of applying them silently
- add focused tests for fix policy, hygiene checks, maintenance behavior, and scheduler execution

## Why
Doctor already had the right abstraction for conservative remediation via `DiagResult.fix`, but most checks were still report-only. This turns doctor into the policy engine for low-risk housekeeping instead of introducing a separate maintenance subsystem.

## Impact
- safe cleanup like temp media, finished eval logs, orphaned artifacts, and eligible session compaction can run automatically from the scheduled hygiene loop
- risky cleanup such as git-backed orphaned workspaces is surfaced as approval-required instead of being auto-applied
- installations that already define their own `scheduler.jobs` will need to add the new `resource-hygiene` job explicitly if they want the default behavior

## Validation
- `npm run format`
- `npm run typecheck`
- `npm run lint`
- `npm exec vitest -- tests/doctor-fix-policy.test.ts tests/doctor.resource-hygiene.test.ts tests/resource-hygiene-maintenance.test.ts tests/scheduler.resource-hygiene.test.ts`

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -490,6 +490,29 @@
   "scheduler": {
     "jobs": [
       {
+        "id": "resource-hygiene",
+        "name": "Resource Hygiene",
+        "description": "Runs conservative doctor-based resource hygiene twice daily and auto-applies only safe cleanup.",
+        "schedule": {
+          "kind": "every",
+          "at": null,
+          "everyMs": 43200000,
+          "expr": null,
+          "tz": "UTC"
+        },
+        "action": {
+          "kind": "system_event",
+          "message": "resource_hygiene_maintenance"
+        },
+        "delivery": {
+          "kind": "last-channel",
+          "channel": "",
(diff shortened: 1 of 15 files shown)


furukama:

PR #316, opened 2026-04-14
[codex] Add mac desktop app

furukama:
## What changed

This adds a new `desktop/` Electron workspace that packages HybridClaw as a native macOS desktop app.

- wrap the existing `/chat`, `/agents`, and `/admin` surfaces inside Electron windows
- bundle the local HybridClaw runtime so the app can reuse an existing gateway or start its own bundled gateway automatically
- add mac-specific window chrome, menu wiring, route handling, and About/help behavior
- add HybridClaw app icon generation, Icon Composer assets, and source-run app bundle support for local development
- add mac packaging support for `.zip` and `.dmg`, including a branded drag-to-install DMG layout
- wire the root workspace scripts, docs, and ignore rules for the desktop workspace

## Why

HybridClaw already has the web surfaces needed for chat and admin flows. This change ships a native mac wrapper around those existing surfaces so the product can be installed and used like a standard mac application.

## User impact

- `npm run desktop` launches a native HybridClaw mac app from source
- `npm run desktop:mac` produces packaged mac artifacts
- the app opens Chat, Agents, and Admin in-app instead of sending those routes to the browser
- packaged builds include the bundled runtime/gateway needed for local use

## Validation

- `npm run format`
- `npm run lint`
- `npm --workspace desktop run test`
- `npm run build:desktop`
- `npm --workspace desktop run dist:mac`


furukama:

maxnoller:
### Code review

Found 3 issues:

1. `.brandWordmark` not hidden in collapsed sidebar state. The PR renames `.brandText` to `.brandWordmark` in JSX and CSS definitions, but the collapsed-state CSS rule (line 313 of `index.module.css`) still targets `.brandText`. After this PR, the "HybridClaw" wordmark will remain visible when the sidebar is collapsed.

https://github.com/HybridAIOne/hybridclaw/blob/e0e5b01af06317ef9e9d7b4f076e820ed34d925a/console/src/components/sidebar/app-sidebar.tsx#L85-L90

2. `shouldCopyEntry` strip logic in `prepare-runtime.mjs` is inverted relative to its comment. The comment says "Only strip when the directory sits directly inside a package" but the condition `!parent.startsWith('@') && parent !== 'node_modules'` strips deeply-nested test dirs (which may be runtime-required) and keeps top-level ones -- the opposite of the stated intent.

https://github.com/HybridAIOne/hybridclaw/blob/e0e5b01af06317ef9e9d7b4f076e820ed34d925a/desktop/scripts/prepare-runtime.mjs#L145-L157

3. `waitForExit` in `gateway-runtime.ts` has a race condition: after the 5-second timeout fires and sends SIGKILL, `stop()` resets `#stopping = false` before the child's `exit` event fires. The exit handler then sees `#stopping === false` and emits `unexpected-exit`, showing a crash dialog for what was actually an intentional stop.

https://github.com/HybridAIOne/hybridclaw/blob/e0e5b01af06317ef9e9d7b4f076e820ed34d925a/desktop/src/gateway-runtime.ts#L76-L85

:robot: Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

maxnoller:
### Sidebar changes — conflict with #323

This PR's sidebar modifications overlap with #323 (`chat-ux-improvements`), which also rewrites the same four files (`app-sidebar.tsx`, `index.module.css`, `index.tsx`, `sidebar.test.tsx`). The two PRs will conflict on merge.

The sidebar changes in this PR serve two independent purposes:

1. **Desktop integration hooks** — the `data-hc-sidebar-root`, `data-hc-sidebar-header`, and `data-hc-main-panel` attributes. These are one-line additive changes that the desktop chrome CSS (`window-chrome.ts`) targets for title bar padding and drag regions. They don't overlap with #323's work.

2. **Branding rework** — swapping the `HybridClaw` React icon for an ``, removing the "Admin console" eyebrow, downgrading `<h1>` to `<span>`, renaming `.brandMark`/`.brandText` to `.brandLogo`/`.brandWordmark`. These changes have issues (collapsed-state rule at line 313 of `index.module.css` still targets the old `.brandText` class, dead `.eyebrow` CSS left behind, `<h1>` removal breaks heading hierarchy) and conflict directly with #323's sidebar work (which adds `.headerRow` layout, toggle button, `collapsible` modes, and its own collapsed-state rules including `.brand { display: none }`).

Suggestion: strip the branding rework from this PR and keep only the `data-hc-*` attributes. The desktop app doesn't depend on the branding changes — the chrome CSS hooks target the `data-hc-*` attributes and existing class names like `.sidebar` and `.topbar`. This lets #316 merge first cleanly, #323 rebases with trivial conflicts (three additive attribute lines), and the branding simplification can land as a small follow-up after #323 that accounts for its collapsed-state handling.

maxnoller (approved):

PR #348, opened 2026-04-17
fix(markdown): restore ordered-list numbering across chat renderers (#320)

### LINKED ISSUE #320: <ol> list formatting web chat broken
Formatting of ordered lists is not working right

Image

maxnoller:
Fixes #320.

## Summary

Ordered-list numbering broke in the web chat (and several other surfaces) whenever:

1. An unordered sub-list sat between numbered items (`ol → ul → ol`) — each numbered item ended up in its own `<ol>` restarting at 1.
2. Sub-bullets were indented under an ordered parent — they broke the `<ol>` instead of nesting inside its `<li>`.
3. The LLM emitted bold-wrapped numbered headings (`**1. Heading**`) — they parsed as bold paragraphs, so numbering came from literal text rather than the `<ol>` counter.

The bug showed up in **three** different markdown renderers because the repo hand-rolls one for the public docs viewer + web chat and wraps `marked` for the console chat. All three are fixed:

- `docs/static/docs.js` + `docs/chat.html`: replaced the flat `openList` string with a stack of `{type, indent, startNum, itemCount, liOpen}` frames, so nested indented lists render as children of their parent `<li>`. Added a `pendingOlStart` pointer so an `<ol>` reopened at the same indent after an interrupting `<ul>` emits `start="N"`, preserving the counter.
- `console/src/lib/markdown.ts`: `marked` was already producing `<ol start="N">`, but the `sanitize-html` allowlist stripped the attribute — added `ol: ['start']`.
- All three now pre-normalize `**N. Heading**` lines to `N. **Heading**` so they parse as real ordered-list items with bold content.

Verified against the actual screenshot input from the issue — numbering now renders sequentially 1–5.

## Why the original fix (#216) didn't catch this

The three regression tests from #216 only covered the *homogeneous* case (`ol→blank→ol`, `ul→blank→ul`, `ol→blank→paragraph`). They never exercised a `<ul>` sitting between two `<ol>` blocks, which is exactly the pattern LLMs produce for headed checklists.

## Test coverage

- `tests/docs-viewer.test.ts`: +22 cases covering headings, inline formatting, external/relative links, code fences, tables, blockquote callouts, HR, XSS escaping, and the three #320 regressions. (29 total, was 7.)
- `console/src/lib/markdown.test.ts`: +18 cases covering nested lists, non-1 start markers, tables with alignment, GFM autolinks, mailto vs. disallowed schemes (ftp/data), stripped ``/`<iframe>`/`<style>` tags, event-handler attribute stripping, and the #320 regressions. (22 total, was 4.)

## Test plan

- [x] `npx vitest run tests/docs-viewer.test.ts` — 29 passed
- [x] `cd console && npx vitest run src/lib/markdown.test.ts` — 22 passed
- [x] Manual verification: running the renderer on the exact screenshot input from #320 produces sequential `<ol>`, `<ol start="2">`, …, `<ol start="5">`.
- [ ] Smoke-test web chat locally after pulling the branch, paste a numbered-list + bullets response and confirm numbering.

## Known follow-up

`docs/chat.html` has no automated test coverage because it's a standalone HTML file with inline `<script>` — the parser was verified by extracting the function and running it through Node directly. A follow-up could extract the parser into a shared ES module imported by both `docs/chat.html` and `docs/static/docs.js`.

### DIFF
--- a/console/src/lib/markdown.test.ts
+++ b/console/src/lib/markdown.test.ts
@@ -143,9 +143,7 @@ describe('renderMarkdown', () => {
   });
 
   it('strips <img>, <iframe>, and <style> tags entirely', () => {
-    expect(renderMarkdown('![alt](https://x.com/y.png)')).not.toContain(
-      '<img',
-    );
+    expect(renderMarkdown('![alt](https://x.com/y.png)')).not.toContain('<img');
     expect(renderMarkdown('<iframe src="x"></iframe>after')).not.toContain(
       '<iframe',
     );
--- a/console/src/lib/markdown.test.ts
+++ b/console/src/lib/markdown.test.ts
@@ -47,4 +47,163 @@ describe('renderMarkdown', () => {
     expect(html).toContain('<ul>');
     expect(html).toContain('<li>first</li>');
   });
+
+  it('returns empty string for empty or whitespace-only input', () => {
+    expect(renderMarkdown('')).toBe('');
+    expect(renderMarkdown('   \n  \n')).toBe('');
+  });
+
+  it('normalizes CRLF line endings', () => {
+    expect(renderMarkdown('line1\r\nline2')).toContain('line1<br />line2');
+  });
+
+  it('renders headings h1 through h6', () => {
+    expect(renderMarkdown('# A')).toContain('<h1>A</h1>');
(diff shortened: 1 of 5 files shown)


maxnoller:
### Code review

PR #370, opened 2026-04-19
[codex] fix pdf layout and artifact path remapping

furukama:
## What changed
- rewrote `skills/pdf/scripts/create_pdf.mjs` to wrap text explicitly with font metrics instead of relying on `drawText(..., maxWidth)` while advancing the cursor by one logical line
- normalized both literal `\n` sequences and actual newlines before layout
- added automatic page creation when wrapped body text exceeds the current page
- added regression tests for newline-heavy wrapped paragraphs and multi-page overflow
- documented the new wrapping and page-break behavior in the PDF skill docs
- updated artifact remapping in the host and container runners to respect custom workspace display roots such as `/app`, not just `/workspace`
- added a regression test for artifact remapping when the runtime reports `/app/output.pdf`

## Why
There were two related user-facing failures:

1. The `/pdf` skill could produce overlapping lines because `pdf-lib` was wrapping long text internally, but the script only moved the vertical cursor once per input line. That let later `drawText()` calls start inside already-rendered wrapped text.
2. Cloud-sandbox artifact downloads could fail even when the file existed in the workspace, because artifact remapping only recognized absolute workspace paths under `/workspace`. In environments that reported artifact paths under `/app`, the host-side remap could not resolve the file back to the real workspace path.

## Impact
- PDFs created through `/pdf` now preserve line breaks, use consistent line height, and continue onto new pages instead of overlapping or truncating long body text.
- Workspace artifacts reported from custom display roots can now be remapped back to the correct host path, which fixes download/attachment handling for those environments.

## Validation
- `./node_modules/.bin/vitest run tests/pdf-skill-node.test.ts tests/pdf-context.test.ts`
- `./node_modules/.bin/vitest run tests/container-runner.artifacts.test.ts`
- `npm run typecheck`
- `npx biome check skills/pdf/scripts/create_pdf.mjs tests/pdf-skill-node.test.ts skills/pdf/SKILL.md`

### DIFF
--- a/skills/pdf/SKILL.md
+++ b/skills/pdf/SKILL.md
@@ -113,6 +113,8 @@ node skills/pdf/scripts/create_pdf.mjs output.pdf --text "Line 1\nLine 2" --font
 For creation tasks ("make a PDF", "create a PDF with X"), always use this bundled
 script or the recipe from [reference.md](./reference.md). Never call `drawText()`
 without passing an embedded `font` — omitting it produces a blank/corrupt page.
+The bundled script wraps long lines, respects explicit `\n` line breaks, and
+adds pages automatically when content exceeds the first page.
 Use a workspace-relative `output.pdf` path for the final deliverable. Reserve
 `/tmp/...` paths for scratch files that do not need to persist after the run.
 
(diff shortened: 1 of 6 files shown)


furukama:

PR #371, opened 2026-04-19
[codex] handle transient transport errors locally

furukama:
## Summary
- add a shared classifier for expected websocket and network transport failures
- add explicit local handlers for both Discord client `error` and Discord `shardError` transport events so reconnect churn is logged instead of bubbling
- keep expected transport exceptions out of the fatal `uncaughtException` exit path and add regression coverage

## Root cause
Normal transport failures such as `Opening handshake has timed out` could escape local handling and reach the global uncaught-exception path, which treated them as process-fatal and exited the gateway. This fix also closes the gap where Discord shard-level transport failures were not handled explicitly.

## Impact
The gateway stays up through transient websocket and network issues while still logging unexpected failures at error or fatal severity.

## Validation
- `npm run format`
- `npm run typecheck`
- `../../node_modules/.bin/vitest run --configLoader runner --config vitest.unit.config.ts tests/transport-errors.test.ts tests/discord.error-handler.test.ts tests/logger.test.ts`

### DIFF
--- a/container/src/index.ts
+++ b/container/src/index.ts
@@ -17,6 +17,7 @@ import { waitForInput, writeOutput } from './ipc.js';
 import { McpClientManager } from './mcp/client-manager.js';
 import { McpConfigWatcher } from './mcp/config-watcher.js';
 import {
+  formatModelErrorForLog,
   isRetryableModelError,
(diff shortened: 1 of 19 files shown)


furukama:

PR #373, opened 2026-04-19
[codex] fix cloud artifact download paths

furukama:
## Summary
- preserve absolute artifact paths when the real workspace already lives under `/workspace`
- keep runtime path resolution from remapping real workspace paths as display-root aliases
- add regressions for the cloud workspace layout under `/workspace/.data/data/agents/main/workspace`

## Why
Cloud container sessions can use a real workspace path under `/workspace/.data/...` while still exposing `/workspace` as the display root. The artifact remapper treated those real absolute paths as display aliases first, which could duplicate the workspace prefix and generate dead `/api/artifact` downloads even though the file existed in the agent workspace.

## Impact
Artifact cards in browser chat keep resolving to the real file path in cloud-container layouts instead of returning `Artifact not found.` for files that were successfully created.

## Validation
- `./node_modules/.bin/vitest run tests/container-runner.artifacts.test.ts tests/container.runtime-paths.test.ts --configLoader runner --config vitest.unit.config.ts`
- `npm run typecheck`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,10 +29,6 @@
   output artifacts are remapped against the active workspace path even when the
   runtime exposes a different display root such as `/app`, keeping generated
   files downloadable and attachable from chat surfaces.
-- **Web chat history search respects authenticated session identity**:
-  `/api/chat/recent` title search now uses the signed session subject for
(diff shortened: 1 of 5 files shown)


furukama:

PR #376, opened 2026-04-20
[codex] Share provider discovery cache lifecycle

furukama:
## Summary
- extract the shared provider discovery cache lifecycle into `createDiscoveryStore` in `src/providers/utils.ts`
- refactor codex, huggingface, hybridai, mistral, openai-compat, and openrouter discovery stores to use the shared helper
- keep provider-specific parsing, logging, and empty-cache behavior intact while making the refactor net negative in LOC

## Why
These provider discovery modules were each reimplementing the same one-hour TTL cache, in-flight request deduplication, and stale-result fallback flow.

## Impact
Provider discovery behavior stays the same, but cache orchestration now lives in one place instead of being copied across multiple providers.

## Root Cause
The discovery cache lifecycle was copy-pasted into several provider-specific modules instead of being centralized behind a shared abstraction.

## Validation
- `npm run typecheck`
- `/Users/bkoehler/src/hybridclaw/node_modules/.bin/vitest run tests/openai-compat-discovery.test.ts tests/mistral-discovery.test.ts tests/hybridai-discovery.test.ts tests/huggingface-discovery.test.ts`

### DIFF
--- a/src/providers/codex-discovery.ts
+++ b/src/providers/codex-discovery.ts
@@ -2,6 +2,7 @@ import {
   getCodexAuthStatus,
   resolveCodexCredentials,
 } from '../auth/codex-auth.js';
+import { logger } from '../logger.js';
 import { CODEX_CLIENT_VERSION } from './codex-constants.js';
 import {
   createDiscoveryStore,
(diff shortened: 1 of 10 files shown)


furukama:

PR #382, opened 2026-04-20
[codex] fix gateway restart shutdown hang

furukama:
## What changed

This fixes two shutdown paths that could make `hybridclaw gateway restart --sandbox host` hang until the CLI fell back to `SIGKILL`.

- Bound the WhatsApp shutdown wait for pending credential writes so a stuck `saveCreds()` cannot block gateway exit forever.
- Changed gateway SIGTERM shutdown ordering so pooled host executors are stopped before the drain loop, which prevents idle cached agent processes from keeping the gateway alive during restart.
- Added a targeted WhatsApp regression test for the stuck credential-save case.

## Why it changed

Gateway restart was timing out even after the HTTP server had begun graceful shutdown.

There were two separate contributors:

1. The WhatsApp connection manager waited indefinitely for `credsSaveQueue` during shutdown.
2. The gateway drain loop waited on `getActiveExecutorCount()`, which includes pooled host executors that remain alive for reuse and are not themselves active work.

Together these could keep the old gateway process alive longer than the CLIs 10 second stop timeout.

## Impact

- Repeated `hybridclaw gateway restart --debug --sandbox host` runs stop cleanly instead of frequently requiring `SIGKILL`.
- Shutdown is more predictable when WhatsApp is enabled and when a recent host executor is still warm in the pool.

## Validation

- `./node_modules/.bin/vitest run tests/whatsapp.connection.test.ts`
- Real restart verification in host sandbox mode against the live runtime:
  - confirmed an older pre-fix process still timed out
  - confirmed subsequent post-fix gateway instances shut down cleanly on repeated restarts

### DIFF
--- a/src/channels/whatsapp/connection.ts
+++ b/src/channels/whatsapp/connection.ts
@@ -343,10 +343,13 @@ async function waitForPendingCredsSave(
   target: WhatsAppLogger,
   pendingSave: Promise<void>,
 ): Promise<void> {
-  let timeoutHandle: ReturnType<typeof setTimeout> | null = null;
+  let timeoutHandle!: ReturnType<typeof setTimeout>;
   try {
     const outcome = await Promise.race([
-      pendingSave.then(() => 'saved' as const),
+      pendingSave.then(
+        () => 'done' as const,
+        () => 'done' as const,
+      ),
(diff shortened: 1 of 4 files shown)


furukama:

PR #397, opened 2026-04-22
ci(release): create GitHub Release only after image & npm publish

maxnoller:
## Summary

Gates GitHub Release creation on successful image + npm publishes, so the Release always corresponds to a pullable version — and creates it as a **draft** so the maintainer's curated release notes (AGENTS.md §7.6 step 8) are never overwritten by auto-generated ones.

### Changes

- New `create-release` job with `needs: [validate, publish-agent, publish-gateway, publish-npm]` that calls `gh release create --draft --verify-tag --generate-notes`. Idempotent: skips if the Release already exists (backwards-compatible with the current UI-first flow). Scoped `contents: write` on this job only; workflow default stays `contents: read`.
- `validate` now emits `prerelease` as a boolean output; `create-release` consumes it via a bash array (safe under `set -u`).
- `publish-npm` preflights `npm view "$NAME@$VERSION" version` and exits 0 if already published, so re-runs after a downstream failure don't 403.

### Operator change

- Today: create Release in UI → tag push triggers workflow → Release already exists when consumers look.
- New (recommended): `git tag vX.Y.Z && git push origin vX.Y.Z` → workflow publishes everything, then creates a **draft** Release with auto-generated notes. Maintainer edits the draft to the AGENTS.md §7.6 curated format (Highlights / Changed / Fixed / Contributors / Full Changelog) and publishes it. Consumers only ever see a published Release whose artifacts are ready.
- Backwards compatible: the UI-first flow still works; `create-release` sees the existing Release (including drafts) and skips.

## Test plan

- [x] YAML parses; `needs` graph is acyclic; `validate` outputs include `prerelease`.
- [x] Bash array `"${flags[@]}"` is safe with empty arrays on bash 5 (GitHub runners).
- [ ] Dry-run via a test tag on a fork: verify the draft Release is created only after all publish jobs succeed, that editing + publishing the draft works, and that re-running against an existing (draft or published) Release no-ops.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -31,6 +31,7 @@ jobs:
       owner_lc: ${{ steps.check.outputs.owner_lc }}
       sha: ${{ steps.check.outputs.sha }}
       sha_short: ${{ steps.check.outputs.sha_short }}
+      prerelease: ${{ steps.check.outputs.prerelease }}
     steps:
       - name: Resolve target ref
         id: resolve
@@ -82,6 +83,10 @@ jobs:
           if [[ -n "${DOCKERHUB_USERNAME}" && -n "${DOCKERHUB_TOKEN}" ]]; then
             PUBLISH_DOCKERHUB=true
           fi
+          PRERELEASE=false
+          if [[ "${TAG}" == *-* ]]; then
+            PRERELEASE=true
+          fi
           OWNER_LC="$(echo "${GITHUB_REPOSITORY_OWNER}" | tr '[:upper:]' '[:lower:]')"
           SHA="$(git rev-parse HEAD)"
           SHA_SHORT="${SHA::7}"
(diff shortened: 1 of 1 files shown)


furukama (approved):

maxnoller:
Worth considering before we merge: the idempotent skip keeps this backwards-compatible with today's UI-first curated-Release flow (AGENTS.md §7.6 step 8), so happy path is fine. But two edges I'd like to close first:

1. **Race vs. the curated Release.** Today `publish-release.yml` never touches Releases, so a slow maintainer is harmless. After this PR, if the curated Release isn't in place before all three publish jobs finish (~3–8 min for multi-arch builds + npm), `create-release` will auto-create with `--generate-notes` — a raw PR/commit list, not the §7.6 curated format (Highlights / Changed / Fixed / Contributors). The maintainer then has to overwrite it.

2. The "Recommended" operator change in the PR description (push tag, let the workflow create the Release) directly contradicts §7.6 step 8's "do not paste the raw CHANGELOG.md version heading/body verbatim."

Suggestion: add `--draft` to the `gh release create` call. That way the workflow never publishes uncurated notes, the race window disappears, and the maintainer just edits + publishes the draft at their own pace — which also preserves the "Release = images pullable" signal as long as consumers look at published (not draft) Releases. If we go that route, the PR description's "Recommended" section can either be dropped or rewritten as "push tag → edit the auto-created draft → publish."

PR #402, opened 2026-04-24
[codex] Improve gateway model response debugging

furukama:
## Summary

- Add provider-boundary prompt dumps and raw model response logging for streaming and non-streaming model calls.
- Add gateway ablation flags for system prompt sections and tool schemas, including `--system-prompt=none` and `--tools=none` / `--no-tools`.
- Stop injecting Qwen chat-template stop sequences into local OpenAI-compatible requests.
- Strip orphan `</think>` markers from normalized visible output when vLLM emits them as content after structured reasoning.

## Why

The gateway needed a way to compare the exact request sent to the model with the raw upstream response. This exposed two issues in vLLM/Qwen debugging: implicit stop sequences were changing streaming behavior, and orphan closing think markers could leak into visible content.

## Validation

- `./node_modules/.bin/vitest tests/thinking-extractor.test.ts tests/local-container-providers.test.ts --run`
- `./node_modules/.bin/vitest tests/cli-flags.test.ts tests/gateway-service.skill-observation.test.ts --run`
- `npx biome check --write container/src/providers/thinking-extractor.ts container/src/providers/local-openai-compat.ts tests/thinking-extractor.test.ts tests/local-container-providers.test.ts`
- `npm run typecheck`
- `npm run lint`


furukama:

PR #407, opened 2026-04-25
feat(session): persist AI-generated session titles

maxnoller:
## Summary

Generate a short title (3-7 words) via the auxiliary model after the first user turn and persist it on the session row. The chat sidebar (`/chat`) and `/api/chat/recent` prefer the stored title, falling back to the existing first/last-message preview when none exists. Pattern mirrors Hermes Agent's `agent/title_generator.py` and DeerFlow's `TitleMiddleware`.

## Why

Today the sidebar always derives a title on read from the first/last message via `buildSessionBoundaryPreview` (`src/session/session-preview.ts`) - clunky labels like `"hey can you" ... "Sure - here's the..."`. A real, persisted title makes sessions scannable and searchable by topic.

## What's in this PR

- **Schema migrations V28/V29** (`src/memory/db.ts`): nullable `title` and `title_source` columns on `sessions`, plus a DB-level constraint so `title_source` is either `NULL` or `'auto'`.
- **DB helpers**: `setSessionTitle(id, title)` writes an automatic title only when the session has no title yet; `getSessionTitle(id)` round-trips the stored title metadata. `getRecentSessionsForUser` prefers the stored title and falls back to the existing boundary preview.
- **Title generator** - new `src/session/session-title.ts`:
  - `generateSessionTitle(...)` - Hermes-style "3-7 words, return only title text" prompt; normalizer strips quotes, `Title:` prefix, `<think>` tags, trailing punctuation; caps stored titles at the shared recent-session title length; rejects empty/single-char/`untitled` outputs.
  - Title input is limited to the first 500 characters of the first user message. Assistant response content is not forwarded to the title model.
  - `maybeAutoTitleSession(...)` - first-turn gate, then `void (async () => ...)()` fire-and-forget body wrapped in `withSpan('hybridclaw.session.title', ...)`. It never blocks the user-visible reply.
- **Auxiliary task wiring**: new `'session_title'` task added to the auxiliary task model config and `config.example.json` (default `provider: "auto"`, override with `AUXILIARY_SESSION_TITLE_MODEL` / `_PROVIDER`). Operators can set `auxiliaryModels.session_title.provider` to `"disabled"` to skip forwarding first-message text to the auxiliary provider.
- **Gateway hook** (`src/gateway/gateway-chat-service.ts`): hooked into concierge respond, version-only, and main success paths using the just-recorded user turn content. The success paths return immediately after scheduling title work, so a first turn enqueues at most one title request.
- **Console**: no changes - `chat-sidebar.tsx:111` already renders `s.title`, which the gateway now resolves to stored-or-derived.

## Out of scope (next iteration)

- `/title` slash command for manual override.
- Backfill for pre-feature sessions (they keep deriving from boundary preview - zero regression).
- Re-titling on long sessions / topic drift.
- Pushing fresh titles to the sidebar in real time (next refresh picks them up).

## Test plan

- [x] `biome check --write` on changed files
- [x] `./node_modules/.bin/vitest run tests/session-title.test.ts tests/memory-service.test.ts tests/providers.task-routing.test.ts`
- [x] `./node_modules/.bin/vitest run --configLoader runner --config vitest.integration.config.ts tests/config-reload.integration.test.ts --passWithNoTests`
- [x] `./node_modules/.bin/vitest run tests/skills-observation.test.ts`
- [x] `./node_modules/.bin/tsc --noEmit`
- [x] `git diff --check`


furukama:

PR #609, opened 2026-04-27
feat(console): model switcher in chat composer

maxnoller:
## Summary

- Adds a portal-based model picker next to the existing agent switcher in the chat composer. Wired to the gateway's `/model set` slash command, hidden from the user-visible thread (`hideUser: true`). Re-seeds from the gateway default whenever the session changes (since `/model set` is session-scoped on the gateway). Surfaces fetch and switch failures via the existing `setError` banner so users get feedback when the dropdown silently snaps back.
- Introduces shared `Select` and `ScrollArea` primitives under `console/src/components/` (portal positioning, viewport flipping, click-outside, typeahead, keyboard navigation with aria-activedescendant, vertical-rail filter, search input slot, item subtitle slot). Built consciously narrow to this picker today; designed so future consumers can adopt the same primitives.
- Extracts a `ChatModel` interface in `console/src/api/types.ts` so the chat surface no longer imports admin-DTO shapes; `AdminModelCatalogEntry` now `extends ChatModel`.
- Model rows show a prettified display name (`claude-haiku-4-5` → **Claude Haiku 4.5**, `gpt-4.1-mini` → **GPT-4.1 Mini**, `deepseek-r1` → **DeepSeek r1**, `o3` stays **o3**) with a quiet subtitle pulled from `parameterSize` / `family` / `vendor`, a *Reasoning* badge when applicable, and a context-window count on the right. Selected row is bolded; provider rail relabels to **Clear filter** when a filter is active.

## Test plan

- [x] `pnpm typecheck` clean from `console/`
- [x] `vitest run` — 242/242 passing (added a `fetchModels` mock to the existing chat-page tests so the new `modelsQuery` resolves; existing assertions unchanged)
- [ ] Manual: open the picker on a fresh session, confirm grouping by provider/vendor, search filters by both raw slug and pretty name
- [ ] Manual: switch sessions and confirm the picker re-seeds to the gateway default rather than carrying the previous session's choice
- [ ] Manual: click a provider in the rail, then click "Clear filter" (renamed "All providers"), confirm filter clears
- [ ] Manual: trigger a switch while a stream is running and confirm the inline error banner appears (`Could not switch model — stop the current run and try again.`)
- [ ] Manual: stop the gateway and reload — confirm the model-load error banner shows (`Failed to load the model list. Model switching is unavailable.`)

## Notes

- The dev server in this worktree wouldn't bring up the SPA cleanly (Vite version mismatch between root and `console/` package), so manual verification was done against the source and against an existing preview build of the prior native `<select>`. A reviewer with a working dev server should run through the manual list above before merge.
- Skipped a wider Composer prop-restructure (switchers array) and a `Dropdown`/`Select` portal-positioning shared hook — both real overlap, but out of scope for this PR; happy to follow up.
- Capability badges beyond `Reasoning` (vision / tool-use / image-gen) are not added because the gateway model catalog doesn't expose those flags yet. The `ChatModel` shape leaves room to add them without further client wiring.


furukama (changes requested):
Die Icons sollten schon die Providerlogos sein.

Image

Und das Dropdown bitte analog zu dem Agent-Chooser daneben. Evtl. das design wiederverwenden?

PR #612, opened 2026-04-27
[codex] Require tokens for gateway API auth

furukama:
## Summary

- Remove loopback-address fallback authentication from gateway API requests when WEB_API_TOKEN is unset.
- Require explicit GATEWAY_API_TOKEN, WEB_API_TOKEN, or existing allowed query-token auth for API access.
- Add regression coverage for loopback requests, forwarded loopback headers, empty bearer tokens, and explicit gateway-token auth.

## Root Cause

hasApiAuth treated the raw socket peer as authenticated when WEB_API_TOKEN was unset. That made local peers bypass bearer-token auth, and it was especially risky when the gateway was deployed behind a proxy or local process boundaries were compromised.

## Validation

- ./node_modules/.bin/vitest tests/gateway-http-server.test.ts --run
- npm run typecheck
- npm run lint
- npx biome check src/gateway/gateway-http-server.ts tests/gateway-http-server.test.ts


furukama:

PR #658, opened 2026-04-29
[codex] Scrub host runtime environment

furukama:
## Summary

- scrub ambient credential-like variables from host agent process environments
- keep web-search provider keys out of process env by passing them through runtime config instead
- redact web-search keys from subsequent IPC input files and prevent host bash calls from inheriting common credential variables

## Why

The host runtime was spawned from the broad parent environment, so LLM-controlled host processes could inherit unrelated credentials such as OpenAI, Anthropic, AWS, and GitHub tokens. This changes the default to deny common secret env patterns while preserving explicit runtime config needed by the agent.

## Validation

- `npm run format`
- `biome check --write` on touched container/test files
- `vitest run tests/host-runner.redaction.test.ts tests/ipc.test.ts tests/unit/web-search.test.ts tests/container.bash-tool.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `git diff --check`

### DIFF
--- a/container/shared/sensitive-env.d.ts
+++ b/container/shared/sensitive-env.d.ts
@@ -1,0 +1,16 @@
+export interface SensitiveEnvRules {
+  exact: ReadonlySet<string>;
+  prefixes: readonly string[];
+  suffixes: readonly string[];
+}
+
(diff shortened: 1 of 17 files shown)


furukama:

PR #660, opened 2026-04-29
[codex] Fix TUI gateway auth and sandbox preflight

furukama:
## Summary

Fixes the local TUI failing to connect cleanly to an already running gateway after loopback API auth was hardened.

## Root Cause

The gateway client could discover a running gateway through unauthenticated `/health`, but the TUI later needed authenticated API calls. When no explicit `WEB_API_TOKEN` or `GATEWAY_API_TOKEN` was configured, the fallback gateway token was generated per process, so a separately launched `hybridclaw tui` could not share the gateway process token.

The TUI sandbox preflight also tried authenticated `/api/status` to learn the running gateway sandbox mode. If auth failed, it fell back to local runtime config and could build the container even when the gateway was already running in host sandbox mode.

## Changes

- Persist the generated fallback `GATEWAY_API_TOKEN` into the encrypted runtime secrets store so local gateway/TUI processes share it.
- Prefer unauthenticated `/health` sandbox diagnostics for TUI preflight before falling back to authenticated `/api/status`.
- Add regressions for host-mode preflight when status auth fails and for generated gateway token persistence.

## Impact

`hybridclaw tui` should now reuse an existing host-mode gateway without rebuilding the container and should authenticate to local gateway APIs without requiring a manually configured token. Existing explicit `GATEWAY_API_TOKEN` and `WEB_API_TOKEN` settings remain authoritative.

## Validation

- `npm run format`
- `./node_modules/.bin/vitest run tests/cli.test.ts tests/runtime-home-layout.test.ts`


furukama:

PR #666, opened 2026-04-29
[codex] secure default localhost web auth

furukama:
## Summary

Adds a secure out-of-the-box localhost auth path for the built-in web console and chat surfaces.

- Issues a process-local HttpOnly `hybridclaw_local_session` cookie for loopback `/chat`, `/admin`, and `/agents` page loads when `WEB_API_TOKEN` is unset.
- Requires either bearer auth or that local cookie for `/api/*`, with same-origin checks on unsafe API methods and admin terminal websocket upgrades.
- Rejects unauthenticated web console access when the gateway is bound to a non-loopback host without `WEB_API_TOKEN`.
- Leaves `/v1/*` OpenAI-compatible bearer-token behavior unchanged.

## Root Cause

The gateway generated a local `GATEWAY_API_TOKEN` by default, so the browser SPA could load from `/chat` but then failed when it called `/api/status` without bearer auth. That broke the documented localhost out-of-box flow.

## Validation

- `npx vitest run tests/gateway-http-server.test.ts`
- `npx vitest run tests/gateway-auth-token.test.ts`
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/src/gateway/auth-token.ts
+++ b/src/gateway/auth-token.ts
@@ -1,10 +1,12 @@
-import { createHmac, timingSafeEqual } from 'node:crypto';
+import { createHmac, randomBytes, timingSafeEqual } from 'node:crypto';
 import fs from 'node:fs';
 import type { IncomingMessage, ServerResponse } from 'node:http';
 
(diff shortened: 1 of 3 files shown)


furukama:

PR #673, opened 2026-04-29
[codex] Decouple gateway health from provider probes

furukama:
## Summary

Decouples the gateway `/health` response from live provider probes so transient HybridAI bot-management outages do not cause Docker liveness checks to hang or fail.

## Changes

- Added `gateway-health-service` to own provider health probe deadline/fallback behavior.
- Made `/health` call `getGatewayStatus({ refreshProviderHealth: false })` so liveness uses cached provider health and skips external refreshes.
- Kept richer status paths able to refresh provider health, with a short deadline and cached fallback for slow HybridAI/local backend probes.
- Added regression coverage for cached HybridAI fallback and the HTTP health endpoint behavior.

## Root Cause

`/health` used the full `getGatewayStatus()` path, which refreshed HybridAI bot-management health. During hybridai.one deploys, `/api/v1/bot-management/bots` can return 502 or time out, making the gateway health check slow or unavailable and causing Docker to mark the instance unhealthy.

## Validation

- `npx biome check src/gateway/gateway-service.ts src/gateway/gateway-http-server.ts src/gateway/gateway-health-service.ts tests/gateway-status.test.ts tests/gateway-http-server.test.ts`
- `npx vitest run tests/gateway-status.test.ts tests/gateway-http-server.test.ts tests/hybridai-health.test.ts`

### DIFF
--- a/src/doctor/checks/providers.ts
+++ b/src/doctor/checks/providers.ts
@@ -5,6 +5,7 @@ import {
 import { getCodexAuthStatus } from '../../auth/codex-auth.js';
 import { getRuntimeConfig } from '../../config/runtime-config.js';
 import { resolveModelProvider } from '../../providers/factory.js';
+import { dedupeStrings } from '../../utils/normalized-strings.js';
 import {
   type ProviderProbeResult,
   probeAnthropic,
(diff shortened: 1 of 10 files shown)


furukama:

PR #675, opened 2026-04-29
[codex] Fix HybridAI model prefix warning

furukama:
## Summary

- Recognize `hybridai/` as an explicit HybridAI provider prefix in the gateway admin model provider resolver.
- Add a regression test for `hybridai/o1-preview` and `hybridai/o3-mini` catalog rows so opening the `/chat` model dropdown does not emit the unknown-prefix warning.

## Root Cause

The admin models path treated bare model slugs as HybridAI, but only listed non-HybridAI provider prefixes in `MODEL_PROVIDER_KEY_BY_PREFIX`. Display/catalog rows with `hybridai/...` therefore fell through to the unknown slash-prefixed warning path even though they are valid HybridAI IDs.

## Validation

- `./node_modules/.bin/vitest --run tests/gateway-status.test.ts -t "hybridai-prefixed models"`
- `npm run typecheck`
- `npm run lint`
- `npm run check`

### DIFF
--- a/src/gateway/gateway-service.ts
+++ b/src/gateway/gateway-service.ts
@@ -248,6 +248,7 @@ import {
   formatModelCountSuffix,
   formatModelForDisplay,
   HYBRIDAI_MODEL_PREFIX,
+  NON_HYBRID_PROVIDER_PREFIXES,
   normalizeHybridAIModelForRuntime,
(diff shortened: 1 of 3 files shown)


furukama:

PR #676, opened 2026-04-29
[codex] Dedupe tunnel reconnect errors

furukama:
## Summary

Fixes duplicate tunnel error text in the console dashboard when a failed reconnect reports the same message that is already persisted as the tunnel `lastError`.

## Root Cause

The public tunnel panel rendered two independent error slots: the persisted tunnel status error and the current reconnect mutation error. When ngrok reconnect failed due to a missing `NGROK_AUTHTOKEN`, both slots contained the same setup message.

## Changes

- Render the reconnect mutation error only when it differs from the persisted tunnel `lastError`.
- Add a dashboard regression test for the ngrok missing-token message so it appears once.

## Validation

- `npm --workspace console run test -- src/routes/dashboard.test.tsx`

### DIFF
--- a/console/src/routes/dashboard.test.tsx
+++ b/console/src/routes/dashboard.test.tsx
@@ -193,7 +193,7 @@ describe('DashboardPage', () => {
         }),
       ),
     );
-    reconnectTunnelMock.mockRejectedValue(new Error(message));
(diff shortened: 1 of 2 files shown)


furukama:

PR #677, opened 2026-04-29
[codex] Add config get command

furukama:
## Summary

Adds a read-only `config get <key>` command for inspecting one existing dotted runtime config value without printing the full config file.

## Changes

- Add a shared runtime config dotted-path getter that reuses the same path validation rules as `config set`.
- Wire `hybridclaw config get <key>` into the local CLI.
- Wire `/config get <key>` through local gateway/TUI command handling and slash command registry metadata.
- Update config help/menu text and focused tests.

## Validation

- `npx vitest run tests/cli.test.ts tests/gateway-status.test.ts tests/tui-slash-command.test.ts tests/tui-slash-menu.test.ts tests/command-registry.test.ts tests/gateway-service.plugins.test.ts`
- `npm run typecheck`
- `npm run format`
- `git diff --check`

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1336,7 +1336,7 @@ async function handleConfigCommand(args: string[]): Promise<void> {
     return;
   }
   if (sub === 'get') {
-    const key = String(normalized[1] || '').trim();
+    const key = (normalized[1] || '').trim();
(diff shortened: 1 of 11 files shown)


furukama:

PR #760, opened 2026-05-01
[codex] Fix repeated TUI tool activity lines

furukama:
## Summary

Fixes repeated transient TUI tool activity output when a long tool preview is repainted by the spinner.

## Root Cause

The TUI activity line could fill the terminal width exactly, so some terminals auto-wrapped during each repaint and left duplicate clipped rows behind. The ANSI truncator also counted Unicode code points as one column and could advance through surrogate pairs incorrectly, which could corrupt the jellyfish emoji into a replacement character.

## Changes

- Add display-cell-aware TUI character width handling for ANSI truncation.
- Format transient tool activity lines one terminal column short to avoid repaint auto-wrap.
- Add regression coverage for a long `bash run shell command ...` preview.

## Validation

- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/tui-format.test.ts`
- `npx biome check --write src/tui.ts tests/tui-format.test.ts`

Note: the focused Vitest run prints a startup warning about `better-sqlite3` because dependencies were installed with lifecycle scripts disabled in this worktree. The formatter test passes.


furukama:

PR #801, opened 2026-05-04
[codex] stabilize chat agent switching

furukama:
## What changed

- Nested `/chat/$sessionId` under `/chat` so creating the first chat session updates route params without remounting the chat page.
- Gated chat page queries behind authenticated `ready` state and seeded the chat status query from the already-validated auth status.
- Gated the context ring query with the same auth readiness check.
- Added regression coverage for the auth bootstrap guard and cleaned up chat session hook test setup.

## Why

The first agent dropdown selection on bare `/chat` created a session and navigated to `/chat/<session>`. Because those routes were siblings, the navigation remounted `ChatPage` while the hidden `/agent switch` command was in flight, making the first selection look like a mini reload and requiring a second selection.

First load also showed 401s for chat-related endpoints when queries fired before the console auth state was ready. The new readiness guard prevents those doomed first requests.

## Validation

- `npm run format` (ran before commit; reverted unrelated formatter-only churn outside this PR scope)
- `npm --workspace console exec -- vitest run --config vitest.config.ts src/routes/chat/chat-page.test.tsx src/routes/chat/use-chat-session.test.ts src/routes/chat/use-chat-stream.test.ts`

### DIFF
--- a/console/src/api/chat-types.ts
+++ b/console/src/api/chat-types.ts
@@ -1,5 +1,3 @@
-import type { AdminCommandResult } from './types';
-
 export interface ChatRecentSession {
   sessionId: string;
   title: string | null;
@@ -140,8 +138,6 @@ export interface MediaUploadResponse {
 export interface BranchResponse {
   sessionId: string;
 }
-
-export type CommandResponse = AdminCommandResult;
 
 export interface ChatMessage {
   id: string;
--- a/console/src/api/chat-types.ts
+++ b/console/src/api/chat-types.ts
@@ -140,8 +140,14 @@ export interface BranchResponse {
 }
 
 export interface CommandResponse {
+  kind?: 'plain' | 'info' | 'error';
   status?: string;
   error?: string;
+  title?: string;
+  text?: string;
+  sessionId?: string;
+  sessionKey?: string;
+  mainSessionKey?: string;
 }
 
 export interface ChatMessage {
--- a/console/src/api/chat-types.ts
+++ b/console/src/api/chat-types.ts
@@ -1,3 +1,5 @@
+import type { AdminCommandResult } from './types';
+
 export interface ChatRecentSession {
   sessionId: string;
   title: string | null;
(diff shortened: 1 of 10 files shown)


furukama:
Updated after manual repro from `/agents -> /chat`:

- Agent/model dropdown switches now call `/api/command` directly instead of sending a hidden slash command through `/api/chat` streaming.
- The command result is appended to the resolved session cache, so when the server maps the generated session key to `sess_...`, the "Session agent set..." message remains visible after navigation.
- Added regression coverage for a bare `/chat` first switch resolving to a server session id.

Re-ran:
- `npx biome check --write console/src/routes/chat/chat-page.tsx console/src/routes/chat/chat-page.test.tsx console/src/api/chat-types.ts`
- `npm --workspace console exec -- vitest run --config vitest.config.ts src/routes/chat/chat-page.test.tsx src/routes/chat/use-chat-session.test.ts src/routes/chat/use-chat-stream.test.ts`

furukama:
Follow-up update: simplified first-session routing as suggested.

- The web client now mints `sess_YYYYMMDD_HHMMSS_<hex>` session ids up front, matching the server's instance-id URL shape.
- A first dropdown switch from bare `/chat` now navigates directly to a stable `sess_...` URL instead of first using an `agent:...` session key and waiting for server correction.
- The switch path waits for the history cache for that session to settle before appending the command result, so the first empty history fetch cannot clobber the visible "Session agent set..." message.

Re-ran:
- `npx biome check --write console/src/lib/chat-helpers.ts console/src/routes/chat/use-chat-session.test.ts console/src/routes/chat/chat-page.test.tsx console/src/routes/chat/chat-page.tsx`
- `npm --workspace console exec -- vitest run --config vitest.config.ts src/routes/chat/chat-page.test.tsx src/routes/chat/use-chat-session.test.ts src/routes/chat/use-chat-stream.test.ts`

furukama:

maxnoller:
### Code review

maxnoller (approved):

PR #804, opened 2026-05-04
fix: agent picker requires only one click after refresh

maxnoller:
## Summary
- The native `<select>` in the chat composer's agent picker was content-sized (~29px wide for "main"), so the pill's 16px padding and the chevron span fell outside its hit area — clicking those did nothing, which read as "needing 2-3 clicks" to actually open the dropdown.
- Overlay the `<select>` across the full pill via `position: absolute; inset: 0; opacity: 0`, and render the visible agent name + chevron beneath it. Any click on the pill now opens the dropdown on first try.
- Extracted a small `agentLabel()` helper to dedupe the `name?.trim() || id` derivation used in both the visible label and the `<option>` list.

## Test plan
- [x] `npm --workspace console run test` (259/259 pass)
- [x] `npm run build:console`
- [x] Manual: load `/chat`, refresh, single click on any part of the agent pill (text, padding, or chevron) opens the dropdown.

### DIFF
--- a/console/src/routes/chat/agent-switch-select.tsx
+++ b/console/src/routes/chat/agent-switch-select.tsx
@@ -5,6 +5,10 @@ export interface AgentSwitchOption {
   name?: string | null;
 }
 
+function agentLabel(agent: AgentSwitchOption): string {
+  return agent.name?.trim() || agent.id;
+}
+
 function ChevronGlyph() {
   return (
     <svg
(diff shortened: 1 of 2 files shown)


furukama:
@maxnoller Hatte ich eigentlich schon gefixt. Problem war, dass /chat keine Session erzeugt hat und deshalb der Agent-Picker erst einmal eine neue Session erzeugt hat (erster Klick)

PR #810, opened 2026-05-05
[codex] Keep gateway alive on transport timeouts

furukama:
## Summary

- Treat expected transport failures that escape local handlers as recoverable process-level warnings.
- Keep fatal process exit behavior for unrelated uncaught exceptions.
- Update logger regression coverage for WebSocket handshake timeout exceptions and transport rejections.

## Root Cause

`Opening handshake has timed out` was already classified as expected network transport churn in channel-specific handlers, but the global uncaught exception fallback still logged it as fatal and called `process.exit(1)` if it escaped those local handlers.

## Impact

Intermittent WebSocket handshake timeouts from gateway transports no longer crash the gateway process. Unexpected application exceptions still fail fast.

## Validation

- `./node_modules/.bin/vitest run tests/logger.test.ts tests/transport-errors.test.ts tests/discord.error-handler.test.ts tests/whatsapp.connection.test.ts`
- `npm run typecheck`
- `npm run lint`
- `git diff --check`

### DIFF
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -14,6 +14,7 @@ import {
   LOGGER_SERIALIZERS,
 } from './logger-format.js';
 import { getTraceContext } from './observability/otel.js';
+import { isExpectedTransportError } from './utils/transport-errors.js';
 
 const VALID_LOG_LEVELS = new Set([
   'fatal',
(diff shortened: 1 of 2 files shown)


furukama:

PR #821, opened 2026-05-05
fix(console): rebuild the / command suggestion panel — visibility, behavior, a11y, UX

maxnoller:
## Summary

What started as two small bugs in the chat composer's `/`-command suggestion panel grew, while exercising it live in Chrome, into a much larger problem: the panel was **not visible to the user at all** in the default layout. This PR fixes that, plus a list of correctness / a11y / UX issues found along the way, and rebuilds the panel on the codebase's existing primitives.

### 1. The panel was invisible (root cause)

The hand-rolled `<div className={css.slashSuggestions}>` rendered with `position: absolute; bottom: 100%`, anchored to the composer card. But `.composerWrapper` (and `.chatMain` / `.chatPage` further up) had `overflow: hidden`, and the panel's box sat *above* the wrapper's bounds — so it was clipped to invisibility. Test users only ever saw a flash of nothing when typing `/`.

Fix: render the panel through the existing `Popover` primitive (`console/src/components/popover`) so it portals to `document.body`, auto-flips above the trigger when there's no room below, and is no longer subject to ancestor clipping. Then build a `SlashSuggestionsPanel` component wrapping the listbox in the existing `ScrollArea` (matching what `Select` already does in the console) so the scrollbar styling is consistent with the rest of the app.

### 2. Double trailing space

Accepting a suggestion whose `insertText` already ended in a space (e.g. `/approve `) appended an extra space, producing `\"/approve  \"` (length 10). Fix: trim trailing whitespace from `insertText` before re-appending exactly one.

### 3. Panel didn't close on outside click

Suggestion items use `onMouseDown` with `preventDefault` (so the textarea doesn't blur), so clicking elsewhere left the panel visible. Fix: covered automatically by `Popover`'s `closeOnOutsideClick`.

### 4. Stale fetches re-opened the panel after dismiss

Closing the panel (Escape / outside-click / send) only flipped state. An already-debounced or in-flight `fetchChatCommands` could resolve later and re-open the panel — most visibly, sending the message cleared the textarea but a late fetch would drop the list onto an empty composer. Fix: a `closePanel()` helper clears the debounce timer and bumps the fetch sequence ref so the in-flight resolve gets discarded.

## A11y wiring (was missing entirely)

- The textarea is now a proper combobox: `role=\"combobox\"`, `aria-autocomplete=\"list\"`, `aria-haspopup=\"listbox\"`, `aria-controls={listboxId}`, `aria-expanded`, `aria-activedescendant` pointing at the active option's stable id.
- Each option has a stable id; the listbox carries that same id and an `aria-label`.
- A polite `aria-live` region announces the result count on every fetch (\"12 commands available\" / \"1 command available\" / \"No commands match /zzz\"), reusing the existing `pluralize()` from `lib/format.ts` and the existing `.srOnly` pattern (lifted into a shared `lib/a11y.module.css`).
- `Home` / `End` / `ArrowUp` / `ArrowDown` (with wrap) navigate; `Tab` / `Enter` accept; `Escape` closes; `mouseenter` syncs the active index so keyboard and mouse can never disagree.

## UX

- **Active option scrolls into view** as you arrow through a long list (was stuck at `scrollTop: 0`).
- **Empty state**: typing `/zzzz` now shows \"No commands match /zzzz\" instead of silently hiding the panel; bare `/` with no results stays closed; Enter on the empty state sends rather than autocompleting.
- **Match highlighting**: the substring of each label that matches the current query is wrapped in `<mark>` (bold + underline-with-offset) so users can see why an item appears.
- **Monospace placeholders**: `<id>`, `[name]`, `<info|list|switch|…>` render in `var(--font-mono)` for syntax legibility.
- **Subcommand indent**: items with `depth >= 2` (e.g. `/agent info`) get a left-pad indent so they read as nested.
- **Trigger after a space**: `/` mid-line (e.g. `hello /clear`) now opens the panel; accepting only replaces the slash-token, leaving surrounding text alone.
- **Touch targets**: items have `min-height: 44px` desktop / 48px under 768px (iOS minimum).
- **Active row contrast**: 2px accent left border in addition to the background tint.

## Architecture

- New `console/src/routes/chat/slash-suggestions-panel.tsx` (panel render, label segmentation, scroll-into-view effect, match highlighting + monospace placeholders).
- New `console/src/routes/chat/slash-context.ts` (pure-text `getSlashContext` helper that locates the slash-token at the caret).
- New `console/src/lib/a11y.module.css` with the shared `.srOnly` class; `sheet/index.module.css` now `composes` from it instead of carrying its own copy.
- Composer keeps state and orchestration; panel handles its own listbox ref and derives option ids from the listbox id.

## Test plan

- [x] `npm --workspace console run typecheck` — clean
- [x] `npm --workspace console run test` — **294/294** (32 new cases across composer + slash-context)
- [x] `npm run check` (biome) — clean
- [x] `npm run build:console` — clean
- [x] Manual verification in Chrome against the local gateway:
  - panel renders above the composer (was clipped, now portaled)
  - `/approve` → `\"/approve \"` (single space)
  - click outside / Escape closes
  - `/conf` highlights the matched substring; placeholders render monospace
  - live region announces \"12 commands available\" / \"No commands match /zzz\"
  - `hello /app` opens the panel mid-line
  - active row has 2px accent border + 53px touch height
  - ArrowDown 12 times keeps the active option in view (scrollTop=416/636)
  - sending a slash-prefixed message after a slow fetch starts no longer reopens the panel after the textarea clears

## Commits

- panel-was-invisible (Popover + ScrollArea refactor)
- double-space + outside-click + escape-on-empty
- combobox/listbox aria wiring + scroll-into-view + Home/End + hover sync + empty state + match highlighting + monospace placeholders + subcommand indent + larger touch targets + accent border + trigger-after-space + mid-line accept + live region
- in-flight invalidation on dismiss / submit (this round of Copilot review feedback)
- code-review pass: reuse `.srOnly` and `pluralize`; derive `liveMessage` in render; collapse `emptyQuery` + `activeQuery`; drop `optionId` prop and the unused `useCallback`; move `listboxRef` into the panel; extract `getSlashContext` to its own file

### DIFF
--- a/console/src/components/popover/index.tsx
+++ b/console/src/components/popover/index.tsx
@@ -86,6 +86,23 @@ export function Popover({
   );
 }
 
+export interface PopoverAnchorProps extends HTMLAttributes<HTMLDivElement> {
+  children: ReactNode;
+}
+
+export function PopoverAnchor({
+  children,
+  className,
+  ...rest
+}: PopoverAnchorProps) {
+  const ctx = usePopoverContext('PopoverAnchor');
+  return (
+    <div ref={ctx.setTriggerEl} className={className} {...rest}>
+      {children}
+    </div>
+  );
+}
+
 export type PopoverAlign = 'start' | 'center' | 'end';
 
 export interface PopoverContentProps extends HTMLAttributes<HTMLDivElement> {
--- a/console/src/components/sheet/index.module.css
+++ b/console/src/components/sheet/index.module.css
@@ -82,13 +82,5 @@
 /* Accessibility helpers */
 
 .srOnly {
-  position: absolute;
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px;
-  overflow: hidden;
-  clip: rect(0, 0, 0, 0);
-  white-space: nowrap;
-  border-width: 0;
+  composes: srOnly from '../../lib/a11y.module.css';
 }
--- a/console/src/components/sheet/index.module.css
+++ b/console/src/components/sheet/index.module.css
@@ -82,5 +82,5 @@
 /* Accessibility helpers */
 
 .srOnly {
-  composes: srOnly from '../../lib/a11y.module.css';
+  composes: srOnly from "../../lib/a11y.module.css";
 }
--- a/console/src/lib/a11y.module.css
+++ b/console/src/lib/a11y.module.css
@@ -1,0 +1,11 @@
+.srOnly {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
+}
--- a/console/src/routes/chat/chat-page.module.css
+++ b/console/src/routes/chat/chat-page.module.css
@@ -1098,6 +1098,7 @@ aside[data-state="collapsed"] .chatSidebarContent {
   justify-content: center;
   gap: 2px;
   padding: 10px 14px;
+  min-width: 0;
   min-height: 44px;
   border-left: 2px solid transparent;
   cursor: pointer;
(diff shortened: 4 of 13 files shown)


maxnoller:
### Code review

furukama (approved):

PR #892, opened 2026-05-08
feat(gateway): proxy Vite dev server through gateway origin

maxnoller:
## Summary

- Adds opt-in `HYBRIDCLAW_DEV_VITE_URL` env var; when set, the gateway proxies SPA HTML, Vite source modules, and the HMR WebSocket to the configured Vite dev server. Page loads under the gateway origin and reuses the existing localhost session cookie, so the dev console hits the same auth path real users do (no `WEB_API_TOKEN` workaround).
- Adds `npm run dev:console` — restarts the gateway with the env var set, then runs Vite in the foreground. One command for HMR-driven console development.
- Documents the workflow in `CONTRIBUTING.md`. Without the env var, gateway behavior is unchanged (still serves prebuilt `console/dist`).

## Why

Currently `npm --workspace console run dev` (Vite on `:4173`) can't authenticate against the gateway: the gateway's `hybridclaw_local_session` cookie is only set for requests it serves itself, and cookies don't cross origins to `:4173`. Devs end up using the prebuilt console at `:9090` (no HMR) or rebuilding `console/dist` repeatedly. Same-origin proxy fixes both.

## Test plan

- [x] `npm run dev:console` → open http://127.0.0.1:9090/chat, page loads under the gateway origin, session works, sent a message and got a reply.
- [x] Verified Vite source modules (`/src/main.tsx`) and `/@vite/client` are 200 via `:9090`.
- [x] `npx tsc --noEmit` passes (pre-existing `@ngrok/ngrok` and `policy-command.ts` biome issues are unrelated, present on `main`).
- [ ] HMR edit-reload cycle (no automated coverage; manual smoke).

## AI assistance

Drafted with Claude Code: design discussion, implementation, and PR text.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,22 @@ npm --prefix container run lint
 npm --prefix container run release:check
 ```
 
+## Console Dev Server
+
+Iterating on the React console with HMR while the gateway handles real auth and
+APIs:
+
+```bash
+npm run dev:console
+```
+
+That starts the gateway daemon with `HYBRIDCLAW_DEV_VITE_URL=http://127.0.0.1:4173`
+then runs Vite in the foreground. Open http://127.0.0.1:9090/chat — the gateway
+proxies SPA HTML, Vite source modules, and the HMR WebSocket to Vite, so the
+page loads under the gateway origin and reuses the existing localhost session
+cookie. Plain `npm run dev` still serves the prebuilt `console/dist`. Stop the
+gateway daemon with `npx tsx src/cli.ts gateway stop`.
+
 ## Validation Expectations
 
 Choose the smallest set of checks that actually covers your change:
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "test:live": "vitest run --configLoader runner --config vitest.live.config.ts --passWithNoTests",
     "setup": "npm --prefix container install",
     "dev": "tsx src/cli.ts gateway",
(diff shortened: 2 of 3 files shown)


maxnoller:
Addressed in 5b46329b:

- **Open-proxy via absolute-form `req.url`** — fixed. Call site now passes parsed `pathname + url.search`; `new URL` always resolves against `DEV_VITE_URL`. Verified with `curl --request-target 'http://evil/src/main.tsx'` — request still went to the configured Vite, never to the spoofed host.
- **Auth gating / loopback restriction** — fixed at two levels: (1) module load refuses `HYBRIDCLAW_DEV_VITE_URL` unless it starts with `http://` AND `HEALTH_HOST` is loopback (logs a warn and disables the passthrough otherwise); (2) per-request `canUseDevViteProxy(req)` also requires a loopback `req.socket.remoteAddress` with no forwarding headers, so reverse-proxied traffic never triggers the dev branch.
- **`proxyViteUpgrade` non-upgrade response** — fixed. Added an `upstreamReq.on('response', ...)` handler that calls `writeUpgradeError(502)` and drains the response, so the client socket doesn't hang if Vite answers with a regular HTTP response.
- **`https:` upstream rejection** — same module-load guard; logs a warn and disables, no runtime crash.

Not changed:
- **Restricting the upgrade fast-path to "looks-like-Vite-HMR"** — the new loopback gate already covers the underlying concern (the only realistic non-Vite upgrade attempt would have to come from a non-loopback origin, which is now refused). Adding a path/header sniff on top would be redundant.
- **Tests** — existing `tests/gateway-http-server.test.ts` mocks heavily; a meaningful test for the proxy would need a real upstream server. Tracking as a follow-up rather than bundling here.

furukama (approved):

PR #893, opened 2026-05-08
fix(console): ignore composer key events during IME composition

maxnoller:
## Summary

The composer's `handleKeyDown` acted on Enter/Tab/Arrow keys regardless of IME composition state. CJK / accented input would have a partial composition cut off and sent as a message (or used to confirm a slash suggestion) instead of being committed by the IME first.

## Fix

Early-return from `handleKeyDown` when `event.nativeEvent.isComposing` is true so the IME owns the keystroke (`composer.tsx:217–219`).

## Test plan

- [x] Three new regression tests in `composer.test.tsx` cover:
  - Enter mid-composition does not call `onSend`.
  - Enter after composition completes still sends.
  - Enter mid-composition while the slash-suggestion panel is open does not select an item.
- [x] All 303 console tests pass with the fix; 2 of the 3 new tests fail without it (verified by reverting locally).

## How surfaced

Found by an exploratory review pass over `console/src/routes/chat/`. Other findings from that pass turned out to be false positives (the optional `messageId` field, and the by-design captured-session-at-send-start streaming behavior — there's an explicit test for the latter at `use-chat-stream.test.ts:510`); only this one was a real bug.

## AI assistance

Drafted with Claude Code: review pass, fix, regression tests, and PR text.

### DIFF
--- a/console/src/routes/chat/composer.test.tsx
+++ b/console/src/routes/chat/composer.test.tsx
@@ -485,5 +485,19 @@ describe('Composer', () => {
       fireEvent.keyDown(textarea, { key: 'Enter', isComposing: true });
       expect(textarea.value).toBe('/');
     });
+
+    it('treats keyCode 229 as composing (Safari/WebKit fallback)', () => {
+      const onSend = vi.fn();
+      renderComposer({ onSend });
+      const textarea = getTextarea();
+      fireEvent.input(textarea, { target: { value: 'こん' } });
+      // Safari fires the confirm-Enter with isComposing: false but keyCode 229.
+      fireEvent.keyDown(textarea, {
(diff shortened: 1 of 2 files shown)


maxnoller:
Addressed in 73e241da:

- **Codex P2 (Safari/WebKit IME Enter fallback)** — added `e.nativeEvent.keyCode === 229` to the guard. Test `treats keyCode 229 as composing (Safari/WebKit fallback)` covers it; verified the test fails without the fallback.
- **Copilot (comment vs. behavior)** — broadened the comment to match the actual behavior ("all keystrokes belong to the IME during composition," not just Enter/Tab/Arrow).

furukama (approved):

PR #901, opened 2026-05-09
Add judge subscriber pattern for skill traces

### LINKED ISSUE #623: [Foundation F11.4] Subscriber pattern for judge-on-trace
Hook so any feature can register: "after every skill run matching X, run the judge with criteria Y, store result in Z."

**Acceptance criteria**
- Subscriber registration API: `registerJudgeSubscriber({ filter, criteria, sink })`.
- Hooks the F2 event bus (`skill_run` subscriber).
- Async / debounced: judge calls don't block the agent's main path.
- Per-subscriber cost accounting (counts toward the consuming feature's budget).

**Depends on:** F11.1, F11.2, F2 (#415).

furukama:
## Summary
- Add `registerJudgeSubscriber({ filter, criteria, sink })` on top of the existing `skill_run` event bus.
- Queue matching skill runs per subscriber and drain judge calls asynchronously after a debounce so event emission does not block the agent path.
- Attribute judge usage to each subscriber's feature budget via `judgeTrace` usage context.

## Tests
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/judge-subscriber.test.ts`
- `./node_modules/.bin/tsc --noEmit --pretty false`

Fixes #623


furukama:

PR #902, opened 2026-05-09
[codex] Gate browser credential fills through SecretRef injection

### LINKED ISSUE #754: [R30.6] F13 wiring — credential fills via keystroke-injection rail
Parent: #700 · Depends on: R30.1 \`BrowserProvider\` interface · F13 #698 ✅

## Scope

Wire F13's \`SecretRef\` keystroke-injection rail into \`BrowserSession.fill()\`. Credentials must reach the DOM **without ever being a string in the LLM context, the trajectory store, or F2 event payloads**.

## Deliverable

- \`fill(selector, value)\` where \`value: SecretRef | string\`:
  - \`string\` path: regular keystroke fill (works for non-secret form fields like search boxes).
  - \`SecretRef\` path: resolves through F13's \`injectIntoElement\` (or equivalent), types into the element handle, never returns cleartext to the calling scope.
- F2 event \`browser.credential_filled\` carries selector + host + skill + SecretRef ID — **no value**.
- F3 ACL on the SecretRef path: per-skill / per-host / per-selector predicate (skill \`invoice-harvester\` may resolve \`cred:datev/*\` only when typing into \`*.datev.de\`).

## Acceptance criteria

- [ ] \`fill()\` accepts \`SecretRef | string\` (type wired in R30.1).
- [ ] SecretRef path goes through F13's keystroke API; cleartext never appears as a string variable in the call chain.
- [ ] Lints / type discipline: \`String(secretRef)\`, template interpolation, JSON-stringify of a SecretRef in skill code is a compile-time or lint error.
- [ ] F2 event \`browser.credential_filled\` emits with selector + host + SecretRef ID + skill — no value.
- [ ] F3 predicate gates which secrets a skill may use on which selectors.
- [ ] Hash-chain audit log entry per credential resolve.
- [ ] Test: a skill that tries to log a SecretRef as a string fails the lint / type check.

## Why a separate child

The runtime cost of \"credentials never appear as strings\" is high enough that it deserves its own review. Bundling with R30.2 would mean the local-mode PR has to argue both \"Playwright works\" and \"the cred type discipline is sound\" in the same diff.

## Out of scope

- F14 2FA wiring (R30.7).
- Provider implementations (R30.2/R30.3/R30.4).

furukama:
## Summary

Closes #754.

Wires `BrowserSession.fill()` so plain strings continue through normal Playwright filling, while `SecretRef` values take a gated credential injection path. Secret fills now require skill context, fail closed unless the active page host is resolvable, evaluate the secret policy against skill/host/selector before resolving, inject through an audited handle path, and emit `browser.credential_filled` with selector, host, skill, and SecretRef metadata only.

Also extends the skill guard with credential-ref stringification checks for `String(...)`, template interpolation, and `JSON.stringify(...)`. The rules cover explicit `secretRef`/`credentialRef` identifiers plus broader credential-looking names such as `secret`, `credential`, `creds`, `password`, and `token`, including simple property access like `creds.token`.

`SecretRef`s are now hardened when they enter audited secret APIs. Hardened refs preserve `source`/`id` metadata for policy checks, but throw on string coercion, template interpolation, and `JSON.stringify(...)`; config-shaped refs remain serializable until they cross into those audited paths.

Browser Use Cloud API keys are kept as `SecretRef` metadata on active sessions and resolved to short-lived `SecretHandle`s per API request instead of being stored as long-lived cleartext strings.

## Boundary note

Playwright ultimately accepts typed text as a string at the browser-driver API boundary. This PR keeps that string materialization inside the audited injection helper, records `secret.unsafe_escape`, disposes the handle immediately, and avoids returning or persisting the cleartext in the caller, trajectory, or audit payload. The skill-stringification discipline is enforced by HybridClaw's skill guard rather than ESLint.

## Validation

- `npm run lint`
- `npm run check`
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/secret-injection.test.ts tests/browser-use-cloud-provider.test.ts tests/browser-credential-fill.test.ts tests/skills-guard.test.ts tests/local-browser-provider.test.ts`


furukama:

furukama:
Addressed Claude review feedback in `7cabea70`:

- Made credential-fill host resolution fail closed when `page.url()` is missing, empty, or invalid.
- Required `url()` on the Playwright fill page surface used by SecretRef fills.
- Renamed the policy check to `assertCredentialFillAllowed` and extracted shared credential-fill params.
- Exported/reused the locator type in provider page typings.
- Removed the no-op `createFinding` wrapper.
- Removed the per-fill `readForInjection` closure and normalized the skill name once for policy/audit use.
- Added a regression test for the fail-closed missing-host case.

Validation:
- `npm run lint`
- `npm run check`
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/local-browser-provider.test.ts tests/browser-use-cloud-provider.test.ts tests/browser-credential-fill.test.ts tests/skills-guard.test.ts`

furukama:
Addressed the AC gap review in `8b594cc4`:

- Browser Use Cloud no longer stores `ActiveCloudSession.apiKey` as a long-lived raw string. Active sessions keep only `apiKeyRef`; each Browser Use Cloud API call resolves a short-lived `SecretHandle` and injects it into the request header through the audited secret-header helper.
- Broadened skill-guard stringification detection beyond literal `secretRef` / `credentialRef` names to cover credential-looking identifiers such as `secret`, `credential`, `creds`, `password`, and `token`, including simple property access like `creds.token`.
- Extended the skill guard test with `String(myCreds)`, template interpolation of `creds.token`, and `JSON.stringify(password)`.
- Updated the PR body to explicitly document the unavoidable Playwright string boundary and that the lint discipline is HybridClaw skill-guard based rather than ESLint.

Validation:
- `npm run lint`
- `npm run check`
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/browser-use-cloud-provider.test.ts tests/skills-guard.test.ts tests/browser-credential-fill.test.ts tests/local-browser-provider.test.ts`

furukama:
Follow-up for the remaining AC notes:

- Added `hardenSecretRef(...)` at audited secret API boundaries. Hardened refs still expose `source`/`id` for policy checks, but throw on string coercion, template interpolation, and `JSON.stringify(...)`, covering the accidental serialization gap without breaking JSON-shaped runtime config before normalization.
- Browser fill and Browser Use Cloud API-key paths now normalize refs through that hardening before resolving secrets.
- The Playwright string boundary remains intentionally documented in the PR body: the string exists only inside the audited helper, emits `secret.unsafe_escape`, and the handle is disposed in `finally`.

Validation rerun:

- `npm run lint`
- `npm run check`
- `npx vitest run --configLoader runner --config vitest.unit.config.ts tests/secret-injection.test.ts tests/browser-use-cloud-provider.test.ts tests/browser-credential-fill.test.ts tests/skills-guard.test.ts tests/local-browser-provider.test.ts`

Issue #829, opened 2026-05-07
[Roadmap R53] Firecrawl scrape + crawl skill


furukama:
## Context

Web ingestion skill wrapping the [Firecrawl](https://www.firecrawl.dev/) API (managed) and the open-source self-host (sovereign path). Distinct from **R30** browser: Firecrawl is HTTP-fetch-and-parse, runs server-side, and is the cheap path for unauthenticated public pages. Where stealth, login, or interaction is needed, **R30** / **R54** Camofox takes over.

## Acceptance criteria

- Firecrawl managed-cloud client adapter (R53.1).
- Self-host adapter (open-source Firecrawl on Docker) (R53.2).
- Skill surface: `scrape.url`, `crawl.site` (R53.3).
- Structured extraction via JSON-schema prompt (R53.4).
- **R47** scoped knowledge ingestion bridge — crawled docs feed agent / skill collections without touching the global corpus (R53.5).
- Per-host crawl policy: rate-limit, robots.txt respect, **F3** host allowlist (R53.6).
- Decision-rule documentation: Firecrawl (HTTP) vs **R30** browser vs **R54** Camofox (R53.7).

## Composes with

- **R21** skill packaging
- **F13** credential rail (Firecrawl API key)
- **R47** scoped knowledge collections (ingestion target)
- **R30** browser substrate (peer for interactive paths)
- **R54** Camofox (peer for stealth paths)

## Out of scope

- Replacing R30 / R54 for authenticated or interaction-heavy crawls.
- Long-running incremental-crawl jobs (would need its own scheduler design).

*Principles I + VII.*

PR #953, opened 2026-05-12
[codex] Add Firecrawl managed skill adapter

### LINKED ISSUE #862: [Roadmap R53.1] Firecrawl client adapter (managed)
Sub-issue of R53.

Client adapter against the managed Firecrawl API: `scrape`, `crawl`, `map`, `extract`. F13 API key.

Depends on: F13 (#698).

furukama:
## Summary

Implements the managed Firecrawl adapter for R53.1 on the same runtime path the bundled skill uses.

- Add a bundled `firecrawl` skill and Node helper for Firecrawl v2 `scrape`, `crawl`, `map`, and `extract` flows.
- Include crawl lifecycle support for status, cancel, and active crawl listing.
- Include extract lifecycle support for start and status/result polling.
- Build gateway `http_request` payloads using `FIRECRAWL_API_KEY` so the API key stays in server-side secret injection and never enters model-visible arguments.
- Keep the implementation single-path: no separate unused TypeScript Firecrawl client remains in `src/`.
- Add the missing R53 roadmap row and sub-issue table.

Closes #862.
Refs #829.

## Validation

- `npx vitest run tests/firecrawl-skill.test.ts`
- `npx biome check skills/firecrawl tests/firecrawl-skill.test.ts docs/content/internal/roadmap.md`
- `npm run typecheck`

## Notes

No live Firecrawl API smoke test was run because this session did not have a real `FIRECRAWL_API_KEY`.


furukama:

furukama:
Addressed the remaining Claude review follow-ups that still apply after the managed TS client was removed:

- Merged `normalizeTargetUrl` / `normalizeTargetUrlPattern` behind an `allowGlob` option.
- Removed the duplicate top-level `credential.bearerSecretName` field; the gateway payload keeps the single `httpRequest.bearerSecretName` value.
- Normalized short command aliases before validating against canonical operations.
- Stopped emitting `costMeasurement` for read-only lifecycle requests with no JSON body.
- Inlined the single-use extraction format helper.
- Parsed `--limit` eagerly using the wide map bound, then preserved the tighter crawl bound in request construction.
- Added targeted tests for the review outcomes above.

The managed-client-specific Claude notes are obsolete because `src/firecrawl/managed-client.ts` and its tests were removed from the PR in the previous cleanup. The long `buildRequest()` chain and CLI `spawnSync` notes are intentionally left as-is: they are non-blocking style/perf caveats, and the current direct branch structure/test pattern is consistent with this small helper scope.

PR #980, opened 2026-05-12
[codex] remove env-backed secret refs

furukama:
## Summary

Removes env-backed SecretRef support so credential bindings resolve only from the stored runtime secret store, while preserving Google OAuth as its explicit runtime provider path.

This also canonicalizes the legacy `browser.browserUseCloud.apiKeyRef` env-shaped default to a stored ref during config load, so existing configs with `{ "source": "env", "id": "BROWSER_USE_API_KEY" }` do not prevent gateway startup after the env SecretRef removal.

## Root Cause

Saved credential defaults could still be represented as env-backed SecretRefs. After removing env secret resolution, existing config files with the old browser-use cloud default were rejected at config load time before migration could repair them.

## Changes

- Removed env-backed SecretRef parsing/resolution and updated policy/source types to store-only.
- Removed default secret-route policy allowance for Google OAuth env token injection, keeping Google OAuth on its dedicated runtime path.
- Updated browser cloud, skill manifest, invoice harvester, docs, and tests to use stored secret refs.
- Added `.nvmrc` and `.node-version` set to Node 22 for workspace/runtime consistency.
- Added regression coverage for legacy browser cloud env-ref canonicalization.

## Validation

- `better-sqlite3` load check under Node `v22.22.2` with ABI `127`
- `npm run build` under Node 22 / npm 11
- `tsc --noEmit`
- `git diff --check`
- Focused Vitest coverage for runtime config SecretRefs, browser credential fill, browser-use cloud provider, local/camofox browser providers, skill manifests, invoice credentials, Google OAuth routes, and secret route CLI/status behavior


furukama:

furukama:
Addressed Claude's actionable review feedback in `643831e4`:

- `normalizeBrowserUseCloudApiKeyRef` now rejects legacy env refs with non-string `id` values using a clear error, and warns when canonicalizing legacy env refs to stored refs.
- `readCredentialStore` no longer falls back to `process.env` when no credential store is provided.
- invoice credential refs with unsupported sources now fail fast with an explicit unsupported-source error.

Validation: `npx vitest run tests/runtime-config.secret-refs.test.ts tests/invoice-harvester.test.ts` passed (61 tests).

PR #985, opened 2026-05-13
Pin-aware chat scrolling with jump-to-latest

maxnoller:
## Summary

Replaces the chat page's per-render `scrollIntoView` auto-scroll with a callback-ref hook (`useStickToBottom`) that owns pin state explicitly. A scroll listener with a 120px threshold flips pinned on/off, a `ResizeObserver` on the message list snaps pinned viewers as content grows, and a hand-rolled rAF ease-out drives the new "↓ Latest" chip's smooth jump (Chrome's native `scrollTo({behavior:'smooth'})` is a no-op on this container's layout).

Sending, regenerating, or approving re-pins so the user lands on the fresh stream. Just scrolling up to read history is preserved while composing. Session switches snap instantly so long histories don't crawl by under a smooth animation.

## Behavior

- **Hydrate**: instant snap to bottom on session load — no crawl through a tall history.
- **User scroll up past 120px**: unpins, "↓ Latest" chip appears anchored above the composer.
- **Stream tokens land while unpinned**: scroll position preserved; you stay where you're reading.
- **Stream tokens land while pinned**: auto-snaps to the new bottom via ResizeObserver.
- **Send / regenerate / approve while unpinned**: re-pins, chip clears, you see your action.
- **Click "↓ Latest"**: 220ms ease-out scroll, pin flips synchronously so the chip hides immediately.
- **Session switch while unpinned**: chip clears (resetToBottom flips pin even when the message area is briefly unmounted during the fetch).

## Non-obvious design notes

- **Callback refs over `RefObject + useEffect`**: the message area is conditionally rendered (empty-state otherwise); the effect would never fire-attach when the element mounts later.
- **Snap-on-attach in `scrollRef`**: children mount before parents, so `contentRef`'s ResizeObserver fires its initial callback before `scrollElRef` is set. The scroll-ref attach callback runs `snapIfPinned()` as a safety belt.
- **Hand-rolled rAF easing in `jumpToBottom`**: native `scrollTo({behavior:'smooth'})` is a no-op on this container. The 220ms ease-out recomputes `target` each frame so it tracks a moving bottom when the user bubble or stream tokens are appended mid-animation.
- **`programmaticScrollRef` flag**: consumes exactly one scroll event after each programmatic write so the listener's threshold check doesn't misread our own auto-snaps as user scrolls.
- **`resetToBottom` flips pin even with no scroller attached**: required for the session-switch path, where `.messageArea` briefly unmounts during the history fetch.


maxnoller:
### Code review

furukama (approved):

PR #988, opened 2026-05-13
[codex] add skill unblock controls

furukama:
## Summary

- Show blocked skills in `/admin/skills` with a red blocked badge and an Unblock action instead of the active/inactive toggle.
- Add unblock support through `/skill unblock <name>`, `hybridclaw skill unblock <name>`, and `/api/admin/skills/unblock`.
- Preserve the scanner block as an explicit reviewed bypass marker on the installed skill copy.
- Include scanner finding details in blocked import errors so TUI users see the exact reason, such as oversized files.

## Validation

- `npx biome check --write src/skills/skills-import.ts tests/skills-import.test.ts`
- `npx vitest run tests/skills-import.test.ts`
- `npm run typecheck`
- Earlier focused checks for this change set: `npx vitest run tests/gateway-service.admin-skills.test.ts tests/gateway-service.skill-command.test.ts`, `npm --prefix console test -- src/api/client.test.ts`


furukama:

PR #1039, opened 2026-05-15
[codex] Add mac CUA browser provider

### LINKED ISSUE #1032: [Roadmap 63] Mac CUA browser-use provider — background-safe operator's-own-browser control
Roadmap entry: **R63** in [docs/content/internal/roadmap.md](https://github.com/HybridAIOne/hybridclaw/blob/main/docs/content/internal/roadmap.md). Composes with **R30** browser track ([#700](https://github.com/HybridAIOne/hybridclaw/issues/700)) and **R43** computer-use track.

## Context

R63 adds a **third browser-control track for macOS** that drives the operator's *own* browser via a computer-use driver instead of via Playwright's CDP path. Three reasons this is its own track and not a Playwright variant:

1. **Logged-in session reuse** — the operator is already authenticated, has extensions installed, has SAML / 1Password / password-manager state mid-flight; spinning up a fresh isolated Chromium throws that away.
2. **Anti-automation pages** — sites that fingerprint `navigator.webdriver` or CDP heuristics fail Playwright but accept human-driven input.
3. **Non-DOM surfaces** — `<canvas>`, video, native plugins, and shadow-DOM-heavy SPAs are pixel-and-input territory, not selector territory.

The differentiator vs. **R30.2** (local Playwright), **R30.3** (managed-cloud pool), **R30.4** (browser-use cloud passthrough), and **R54** (Camofox stealth-Playwright) is that R63 controls *the operator's existing browser process* without raising it, warping their cursor, or switching their active Space.

### Driver layer

- Bundled **cua-driver** binary (open-source upstream), invoked over **MCP stdio** by the gateway's existing MCP transport.
- Lazy-installed via `hybridclaw doctor cua-mac` — Mac-only `which cua-driver` check, fetch + install upstream script when absent, version-pinned via `HYBRIDAI_CUA_DRIVER_VERSION` env var.
- New module `src/browser/mac-cua-provider.ts` with `MacCuaBrowserProvider implements BrowserProvider` (per R30.1).

### Permission model

- macOS TCC grants required: **Accessibility** + **Screen Recording**, both granted to the terminal / app running HybridClaw (not the driver binary).
- `hybridclaw doctor cua-mac` detects missing grants and prints the deep-link System Settings URLs:
  - `x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility`
  - `x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture`
- Provider **refuses to advertise into the F5 capability matrix** until both grants land — a skill author never picks R63 in a config that can't run it.

### Action vocabulary (BrowserProvider parity)

`goto`, `click`, `type`, `screenshot`, `fill`, `selectOption`, `scrollIntoView` map onto CUA primitives:

| BrowserProvider action | CUA mapping |
|---|---|
| `goto(url)` | focus address bar via `cmd+L` → `type_text_chars(url)` → press `return` |
| `click(selector)` | resolve selector → AX element → `click(x, y)` at element bounds |
| `type(text)` | `type_text_chars(text)` against focused AX element |
| `fill(selector, value)` | `click(element)` → `type_text_chars(value)` |
| `selectOption(selector, value)` | `click(element)` → `set_value(element, value)` against `AXPopUpButton` |
| `screenshot(mode)` | three modes: `som` / `vision` / `ax` (see below) |
| `scrollIntoView(selector)` | resolve AX bounds → `scroll(direction, amount)` until in viewport |

### AX-tree-first targeting

- Element index + window_id is the **preferred** path — deterministic across retina / HiDPI / browser zoom.
- Raw (x, y) pixel targeting is the **fallback** when AX doesn't expose the target.
- Every pixel-fallback emits an F2 event so we can observe how often it fires.

### Browser-specific affordances

- Target browsers by bundle ID: `com.apple.Safari`, `com.google.Chrome`, `org.mozilla.firefox`, `com.brave.Browser`, `company.thebrowser.Browser`.
- Tab control: `cmd+1`..`cmd+9` (jump to tab N), `cmd+T` (new), `cmd+W` (close), `cmd+shift+]` / `cmd+shift+[` (next / prev tab).
- URL bar: `cmd+L`.
- History: `cmd+[` / `cmd+]` (back / forward).
- Browser identity advertised in the session-lease record so audit logs distinguish which browser drove the action.

### Background-safe mode (load-bearing invariant)

- **Never raise the window.** Input is posted directly to the browser's PID via the driver's event-routing path.
- **Never warp the cursor.** Operator's pointer position is untouched.
- **Never switch Spaces.** The browser can be on a different Space and still receive input.
- **No foreground-stealing fallback at v1.** When AX-based posting fails on a particular window (some Electron shells under-expose AX), the action returns an error rather than disrupting the operator.

### Screenshot path — three modes

| Mode | Output | Use case |
|---|---|---|
| `som` | screenshot + AX tree + numbered element-overlay | vision-capable model picks elements by index |
| `vision` | raw JPEG/PNG | pixel-space model that reasons over the image alone |
| `ax` | AX tree only, no image | model can read structure cheaper than pixels |

- Retina-correct via the driver's logical-pixel coordinate space (no 2x math at the call site).
- Multi-display via global screen-space origin (coordinates can be negative on mirrored / extended displays).
- Output base64-encoded; format detected from magic bytes.

### Safety guardrails (provider-level, before F8)

**Hard-blocked key combos** — refused unconditionally, emit F2 audit event, not stakes-escalatable:
- `cmd+shift+Q` (logout)
- `cmd+shift+delete` (empty trash)
- `cmd+ctrl+Q` (lock screen)
- `cmd+option+shift+Q` (force logout)

**Hard-blocked `type` patterns** — shell-injection refused unconditionally:
- `curl ... | bash`, `curl ... | sh`, `wget ... | bash`
- `sudo rm -rf`
- Fork-bomb shapes (`:(){:|:&};:`)

Reasoning: the model has no plausible business issuing these inside a browser window. They're not stakes-escalatable — they're rejected at the provider.

## Acceptance criteria

- [ ] `MacCuaBrowserProvider` implements the R30.1 `BrowserProvider` interface; smoke tests on Safari + Chrome at minimum.
- [ ] `hybridclaw doctor cua-mac` detects missing TCC grants (Accessibility, Screen Recording) and refuses to advertise the provider into **F5** until both granted.
- [ ] **Background-safe contract test**: a 60-second drive-the-browser action sequence asserts cursor position + frontmost-window + active-Space are all unchanged at the end.
- [ ] Every action emits **F2** events readable from the **R30.9** audit surface (`browser.session_started`, `browser.action`, `browser.screenshot_taken`, `browser.session_ended`).
- [ ] Hard-blocked key combos + shell-injection patterns reject deterministically — regression tests for each.
- [ ] **F13 `SecretRef`** round-trips through `type_text_chars` without cleartext crossing the model boundary — regression test.
- [ ] AX-tree-first targeting is the preferred path — pixel fallback fires only when the resolved element has no usable AX bounds, captured in an F2 event.
- [ ] **R30.6** F13 keystroke-injection parity: when a `fill()` call carries a `SecretRef`, the provider injects the cleartext at the driver boundary, not before.
- [ ] **R30.7** F14 2FA handover hook: when the AX tree shows a 2FA modal or device-check waypoint, the provider parks to F14 the same way the DOM track does.
- [ ] **R30.8** per-host F3 policy: URL extracted from the address-bar AX value is checked against the allowlist *before* the `return` keystroke is dispatched — denied URLs never complete navigation.

## Anti-goals

- **Not** a generic OS-control provider — that's R43 proper. R63 is the browser-window slice with browser-specific affordances baked in.
- **Not** a Linux / Windows path. cua-driver depends on macOS-private APIs; cross-platform CUA is a separate roadmap row.
- **Not** a Playwright replacement. R30.2 / R30.3 / R30.4 remain the right path when a *fresh isolated* browser session is what's wanted — R63 is for the operator's *existing* browser.
- **Not** a screen-scraper for non-browser apps. Use R43 for those.
- **Not** a recording / replay surface. The audit trail is the artifact.
- **Not** a foreground-stealing fallback. When AX posting fails, the action errors out rather than disrupting the operator.

## Composes with

- **R30.1** [#749](https://github.com/HybridAIOne/hybridclaw/issues/749) — `BrowserProvider` interface, same vocabulary, different transport
- **R30.6** [#754](https://github.com/HybridAIOne/hybridclaw/issues/754) — F13 keystroke-injection, opaque-handle discipline
- **R30.7** — F14 2FA handover, same hook points
- **R30.8** — per-host policy, enforced before `return` on URL navigation
- **R30.9** — session audit + screenshot trail, same F2 event vocabulary
- **R43** — computer-use track parent, R63 is the first concrete consumer
- **R54** — Camofox stealth-Playwright sibling, four tracks behind one interface
- **F3** — host allowlist enforced via address-bar AX-value extraction
- **F5** — capability matrix, provider refuses to advertise until TCC grants land
- **F13** — `SecretRef` opaque handles cross the driver boundary, cleartext never reaches the model
- **F14** — escalation modality for 2FA / captcha / device-check waypoints
- **F8** — stakes gating applies *after* the provider's hard-block list; the hard-blocks are unconditional refusals

## Out of scope for v1

- AppleScript / Shortcuts integration (separate follow-up if browser-specific automation gaps appear).
- Safari extension companion (R63 controls the browser from outside; an extension-side companion is a different shape).
- Multi-monitor cursor warping (not needed since v1 doesn't warp at all).
- Macro recording for later replay (operator workflows live in **R2** YAML workflows, not in CUA traces).
- Drag-and-drop actions (the underlying driver doesn't support drag at this layer; falls back to error).

furukama:
## Summary

Implements the macOS CUA browser-provider work for #1032.

Closes #1032

- Adds `MacCuaBrowserProvider` with Safari, Chrome, Firefox, Brave, and Arc bundle IDs, an injectable/stdin driver adapter, AX target parsing, screenshot modes, background-safe checks, and standard browser audit events.
- Wires `mac-cua` into runtime browser config, the provider factory, `config.example.json`, and browser configuration docs.
- Adds `hybridclaw doctor cua-mac` readiness checks for the CUA driver and macOS Accessibility / Screen Recording grants.
- Adds F13 SecretRef policy/audit gating before opaque driver handoff, F14 waypoint audit hooks, address-bar AX validation before navigation commit, real-driver capability gating, stable audit run IDs, stdio driver call timeouts, required background-safety state probes, constructor-time platform fail-fast, and destructive browser shortcut blocking.
- Adds AX-first target resolution through the driver; pixel fallback is only accepted from AX resolution and emits a `browser.pixel_fallback` audit event.
- Covers acceptance behavior with tests for Safari + Chrome smoke launch, SecretRef passthrough, unsafe typed-payload blocking, destructive key chord blocking, AX target handling, pixel fallback audit, provider factory selection, doctor readiness messaging, standard audit event names, address-bar validation, F14 waypoint events, capability advertisement, and the simulated 60-second background-safe contract.

## Validation

- `npx vitest run tests/mac-cua-provider.test.ts tests/browser-provider-factory.test.ts tests/doctor-cua-mac.test.ts tests/managed-cloud-provider.test.ts`
- `npm run lint`
- `npm run check`

## Notes

This workspace does not have a real `cua-driver` / macOS TCC grant flow available, so Safari and Chrome smoke coverage is through the injectable driver contract rather than live browser automation.


furukama:

PR #1042, opened 2026-05-15
[codex] Fix aux local routing and HeyGen response handling

furukama:
## Summary

- Prefer healthy local auxiliary backends before remote fallback/session models for auto-routed aux calls, including CV narration.
- Return truncated `http_request` response bodies with metadata instead of failing oversized API responses outright, while short-circuiting declared over-limit `Content-Length` responses without streaming the body.
- Increase HeyGen asset discovery response budget for avatar and voice listing requests.
- Harden the HeyGen CJS helper with default summarized asset-list output, explicit truncated-response failures, bounded `--watch` status polling, and local asset-id cache validation.
- Simplify HeyGen helper internals after Claude review: flat asset-list extraction, one request build per `request` command, no single-use non-negative parser, and explicit `--skip-cache-validation` coverage.
- Add video-from-script guidance for the packaged hyphenated skill path and exact HeyGen asset ids, including a narrow `--skip-cache-validation` escape hatch for known private assets.
- Add web chat MP4/video artifact previews with a `<video controls>` player and inline video MIME handling.
- Return generated videos from `video.from-script` as `artifacts[]` so web chat can show an artifact card and browser-accessible artifact route.
- Recover generated image/video artifacts at the gateway when the assistant leaks a `.generated-images/...` or `.generated-videos/...` path in final text, matching the generated-image artifact behavior instead of trusting prompt compliance.
- Preserve assistant text when recovering generated media artifacts, so web chat shows response text plus embedded image/video media for both images and videos.
- Tighten generic deliverable prompt wording so web chat returns image/PDF/video artifacts in the final response, uses artifact cards for generated images and videos, never returns only a host-local workspace path, never hand-builds relative `/api/artifact` links, and avoids the `message` tool unless a file-capable external channel is explicitly requested.
- Add explicit recovery guidance for already-generated images/videos: rerun the artifact-producing helper or status/download command so the final response carries `artifacts[]`, and do not claim web chat cannot embed or offer Finder/Discord/email fallbacks.
- Contain CV narration failures with a clearer narration-specific warning and longer timeout.

## Root Cause

Auto auxiliary routing could choose a remote catalog fallback before a reachable local backend, so CV narration tried a remote OpenAI-compatible path even when local vLLM should have been preferred. Separately, the gateway treated oversized outbound API responses as hard errors, which made large HeyGen list responses unusable. The HeyGen skill also encouraged raw `http_request` usage for large list and polling workflows instead of keeping those lifecycle details inside the helper.

The later HeyGen session exposed four model traps: runtime workspaces can mount `video.from-script` as `skills/video-from-script`, avatar/voice display names are easy to confuse with HeyGen's opaque ids, generic deliverable instructions said to "return/upload/post" and "prefer posting" assets even though the message-tool prompt said not to use `message` for web chat artifacts, and host-local or relative workspace paths are not usable by browser users. The helper now caches summarized asset lists and rejects display names or stale cached ids before contacting HeyGen. The web chat now previews video artifacts directly, generated image artifacts already follow the artifact-card path, `video.from-script` emits artifact metadata for the gateway, and the gateway now recovers generated image/video artifacts from leaked final-text paths as a hard runtime fallback while preserving the assistant response.

## Claude Review Follow-up

- Removed the `resolvePreferredLocalTextCallContext` wrapper and added a reachable-local-backend guard before local aux candidate resolution.
- Reused the parsed HeyGen request in `executeRequestCommand` instead of rebuilding it for validation and execution.
- Replaced recursive HeyGen asset extraction with direct `json`/`json.data` lookup plus a shallow top-level fallback.
- Removed `parseNonNegativeInteger` by adding a `min` option to `parsePositiveInteger`, and inlined the single-use asset-kind helper.
- Centralized generated media MIME lookup in `gateway-utils.ts`.
- Restored fail-fast `Content-Length` truncation behavior for outbound `http_request` responses.
- Added an explicit `--skip-cache-validation` regression test.

## Validation

- `npx vitest run tests/providers.auxiliary.test.ts tests/agent-cv.test.ts`
- `npx vitest run tests/gateway-http-server.test.ts --testNamePattern "http_request responses"`
- `npm --workspace console run test -- message-block.test.tsx`
- `npx vitest run tests/gateway-http-server.test.ts --testNamePattern "video artifacts inline|uploaded-media-cache artifacts|active artifact types"`
- `npx vitest run tests/video-from-script-skill.test.ts tests/prompt-hooks.tool-summary.test.ts`
- `npx vitest run tests/heygen-skill.test.ts tests/providers.auxiliary.test.ts`
- `npx vitest run tests/gateway-http-server.test.ts --testNamePattern "http_request responses|content-length exceeds|video artifacts inline|uploaded-media-cache artifacts|active artifact types"`
- `npx vitest run tests/gateway-prompt-env-defaults.test.ts`
- `node skills/heygen/heygen.cjs eval-scenarios`
- `node skills/video.from-script/video-from-script.cjs --help`
- `npx biome check --write src/providers/auxiliary.ts src/gateway/gateway-utils.ts src/gateway/gateway-http-server.ts src/gateway/gateway-chat-service.ts src/gateway/gateway-http-proxy.ts tests/gateway-http-server.test.ts tests/heygen-skill.test.ts skills/heygen/heygen.cjs skills/heygen/client.cjs`
- `npx biome check --write src/gateway/gateway-chat-service.ts tests/gateway-prompt-env-defaults.test.ts`
- `npx biome check --write console/src/routes/chat/message-block.tsx console/src/routes/chat/message-block.test.tsx console/src/routes/chat/chat-page.module.css src/gateway/gateway-http-server.ts src/agent/prompt-hooks.ts tests/gateway-http-server.test.ts tests/prompt-hooks.tool-summary.test.ts skills/video.from-script/video-from-script.cjs tests/video-from-script-skill.test.ts`
- `npx biome check --write src/agent/prompt-hooks.ts tests/prompt-hooks.tool-summary.test.ts`
- `npm run format`
- `npm run typecheck`
- `npm run lint`


furukama:

PR #948, opened 2026-05-12
[codex] Harden npm supply-chain controls

furukama:
## Summary

- Require npm 11.10.x for root and container installs so npm's release-age policy is enforced consistently.
- Add a 7-day npm `min-release-age` gate, CI npm signature verification, and documented dependency update workflow.
- Remove unused direct root dependencies on `@opentelemetry/semantic-conventions` and `html-to-text` while leaving transitive copies resolved by their actual dependents.
- Keep release publishing on the pinned npm CLI while preserving provenance publishing.

## Security Notes

The repo still has a pinned GitHub dependency through `@whiskeysockets/baileys` -> `libsignal`, so this PR documents that as the blocker before enabling npm's `allow-git` restriction.

## Validation

- `git diff --check`
- JSON parse check for root and container package manifests/lockfiles
- `npm ci --dry-run --no-audit --fund=false` with npm 11.10.0
- Earlier local `npm audit signatures` checks passed before the unused direct dependency removal: root verified 1478 registry signatures / 191 attestations, container verified 361 registry signatures / 29 attestations.

PR #1058, opened 2026-05-21
[codex] Harden dependency supply chain

furukama:
## Summary

- Pin remaining direct npm dependency ranges and ship root/container npm shrinkwraps for published transitive dependency resolution.
- Add dependency policy checks for exact direct deps, lockfile review, and newly introduced dependency lifecycle scripts.
- Disable lifecycle scripts for self-update installs and packaged container dependency bootstrapping, then run HybridClaw's container bootstrap explicitly.
- Add scheduled npm audit plus registry signature verification in GitHub Actions.

## Impact

This aligns HybridClaw's npm supply-chain posture with the stricter model discussed: direct dependencies stay exact, published installs carry shrinkwrapped transitive deps, lockfile changes need explicit review, new lifecycle scripts need explicit review, and recurring registry checks surface vulnerable or unsigned dependency state.

## Validation

- `npm run deps:policy`
- `npx biome check scripts/check-dependency-policy.mjs scripts/postinstall-container.mjs scripts/release-check.mjs src/update.ts tests/postinstall-container.test.ts tests/update.test.ts`
- `npx vitest run --configLoader runner --project unit tests/postinstall-container.test.ts tests/update.test.ts`


furukama:

maxnoller:
### Code review

Found 3 issues:

1. `checkStagedLockfiles` returns early after `fail()`, skipping the new-lifecycle-script detection. Because `fail()` only sets `process.exitCode` and the function `return`s right after, a commit that simultaneously introduces a lockfile change AND a new `hasInstallScript` dependency only reports the lockfile-block error. The developer sets `HYBRIDCLAW_ALLOW_LOCKFILE_CHANGES=1`, re-runs, and only then discovers the lifecycle-script warning — a two-pass flow that risks the lifecycle gate being ignored on the first reflexive override.

https://github.com/HybridAIOne/hybridclaw/blob/430f5949366ae957b0830d31ae723cff736629ab/scripts/check-dependency-policy.mjs#L162-L195

2. `deps:update-lockfile` only regenerates `package-lock.json`, but this PR introduces `npm-shrinkwrap.json` (and `container/npm-shrinkwrap.json`) as the published lockfiles. `npm ci` prefers shrinkwrap over package-lock, so any routine `npm run deps:update-lockfile` will leave shrinkwrap stale, silently diverging CI/published installs from the developer's package-lock until someone manually runs `npm shrinkwrap`. The new `check-dependency-policy.mjs` does not detect the mismatch either.

https://github.com/HybridAIOne/hybridclaw/blob/430f5949366ae957b0830d31ae723cff736629ab/package.json#L78-L82

3. `container/Dockerfile` still runs `npm install --global npm@11.10.0 --no-audit --fund=false` without `--ignore-scripts` in both the builder and runtime stages. The PR adds `--ignore-scripts` to the equivalent steps in `setup-node-workspace/action.yml`, `dependency-audit.yml`, and `publish-release.yml`, so the Docker image build is now the only path that still executes npm's bootstrap lifecycle scripts — an inconsistent posture given the PR's stated supply-chain hardening goal (and a gap previously flagged on #948).

https://github.com/HybridAIOne/hybridclaw/blob/430f5949366ae957b0830d31ae723cff736629ab/container/Dockerfile#L4-L7

https://github.com/HybridAIOne/hybridclaw/blob/430f5949366ae957b0830d31ae723cff736629ab/container/Dockerfile#L27-L30

src/update.ts
```diff
@@ -440,6 +514,9 @@ export async function runUpdateCommand(
   if (exitCode !== 0) {
     throw new Error(`Update command failed with exit code ${exitCode}.`);
   }
+  runExplicitPostinstall(
```
maxnoller: This self-update path installs HybridClaw with `--ignore-scripts`, then only runs the container bootstrap script. That leaves reviewed root runtime dependencies with install/build lifecycle steps skipped during global updates; the current shrinkwrap includes root packages such as `better-sqlite3`, `node-pty`, and `onnxruntime-node` with `hasInstallScript: true`. A published update can therefore leave the CLI installed without required native bindings even though the update command exits successfully. Please add an explicit reviewed rebuild/bootstrap step for the required root native dependencies, similar to the CI `npm rebuild better-sqlite3 --ignore-scripts=false` pattern, or otherwise prove those root lifecycle scripts are not needed for packaged global installs.

.husky/pre-commit
```diff
@@ -1 +1,2 @@
+node ./scripts/check-dependency-policy.mjs --staged
```
maxnoller, line 1: The lockfile and new lifecycle-script review gate is currently enforced only through the Husky `--staged` path. That is useful locally, but it is bypassable via `--no-verify`, web edits, automation, or any CI path that runs `check-dependency-policy.mjs` without `--staged`. Since this PR is introducing the gate as a supply-chain control, please add a CI/PR check that compares lockfiles against the merge base and fails on unapproved lockfile changes or newly introduced `hasInstallScript` packages.

maxnoller (changes requested):
I found two remaining issues in the current PR diff that should be addressed before merge. This review is scoped to the actual PR delta only.

maxnoller (approved):

PR #1066, opened 2026-05-21
[codex] Add admin secret mutation API

### LINKED ISSUE #1047: R59.2: Admin secrets overwrite + unset API
## Summary

Add write-only endpoints to overwrite or unset a secret in the existing runtime secret store, so operators can rotate or remove secrets from the admin console (R59.4) without using the CLI.

## Problem

Operators today must shell into the host to rotate or remove a secret via `hybridclaw secret set / unset` (`src/cli/secret-command.ts`). The admin console has no equivalent write path. The challenge is exposing write actions while preserving the no-read-back contract: a cleartext value crosses the boundary on the way in but is never returned on the way out.

## Proposed shape

- `PUT /api/admin/secrets/{name}` — body contains the new cleartext value; response contains only the post-write metadata (fingerprint, length, `last_rotated_at`). Never echoes the submitted value.
- `DELETE /api/admin/secrets/{name}` — unsets the secret; idempotent on already-unset entries.
- Thin wrapper over `runtime-secrets.ts`; no GET counterpart on this route — `GET` returns 405 by design to make the no-read-back contract self-documenting.
- Gated by R14 RBAC action `secret.overwrite` / `secret.unset` (R59.5); overwrite is strictly higher privilege than `secret.list_metadata`.
- Emits F2 `secret.overwritten` / `secret.unset` audit events (R59.6) with operator identity, source IP, and the new-value fingerprint — never cleartext.
- Request body is scrubbed from logs and error traces.

## Acceptance criteria

- Overwrite stores the new value and updates `last_rotated_at`; response carries only metadata.
- Unset removes the value but preserves the entry if declared by a manifest/route (state flips to `unset`).
- `GET /api/admin/secrets/{name}` returns 405; the message does not leak whether the name exists.
- Unauthorized callers (no `secret.overwrite`) get 403 without the value being persisted or logged.
- Audit events are emitted on both success and failure paths; failures do not leak cleartext into the audit chain.
- Tests cover: overwrite, unset, idempotent unset, RBAC denial, value scrubbing from logs, 405 on GET.

## Scope boundaries

- No bulk import/export in v1.
- No time-bounded reveal / break-glass debug flow.
- No OAuth issuance — that belongs on per-connector setup pages and R42.
- No cron-driven rotation — that composes with R41 Autopilot as a follow-up.

## Roadmap

R59.2 in the internal roadmap.

furukama:
## Summary

- Add write-only `PUT /api/admin/secrets/{name}` and `DELETE /api/admin/secrets/{name}` handlers for runtime secret rotation and removal.
- Return post-write metadata only, preserving the no-read-back contract and rejecting `GET /api/admin/secrets/{name}` with a non-enumerating 405.
- Add admin action gating for `secret.list_metadata`, `secret.overwrite`, and `secret.unset`, plus scrubbed success/failure audit events for mutation attempts.

## Validation

- `npx vitest run tests/gateway-admin-secrets.test.ts tests/gateway-http-server.test.ts`
- `npm run lint`
- `git diff --check`

Closes #1047


furukama:

PR #1116, opened 2026-05-27
Fix provider prefix dispatch

### LINKED ISSUE #1112: Refactor provider resolution to use a prefix registry map
## Context

`resolveProviderForModel` currently lives in `src/providers/factory.ts` and resolves by rebuilding the active provider list, then linearly searching `provider.matchesModel(normalizedModel)`. Because `hybridAIProvider.matchesModel` is the catch-all and the resolver also falls back to `hybridAIProvider`, typoed slash-bearing model ids can silently route to HybridAI instead of failing or surfacing a bad provider prefix.

This was requested for the resolver originally referenced as `src/gateway/factory.ts:75`; in the current checkout the relevant code is `src/providers/factory.ts` around `resolveProviderForModel`.

Pi has a cleaner registry shape in `~/examples/pi/packages/ai/src/api-registry.ts`: `apiProviderRegistry = new Map<string, RegisteredApiProvider>()` with explicit registration and direct lookup helpers. HybridClaw can mirror that pattern for provider-prefix dispatch.

## Proposal

Replace the linear provider search with an explicit prefix-to-provider/handler `Map`:

- Register known provider prefixes, e.g. `openai-codex/`, `anthropic/`, `openrouter/`, local backend prefixes, and enabled OpenAI-compatible providers.
- Resolve slash-bearing model ids by extracting the provider prefix and doing a direct map lookup.
- Preserve the existing convention that bare model slugs are HybridAI passthroughs.
- Preserve local model metadata fallback where a bare local model is discovered by backend, if still required by current local-provider behavior.
- For unknown slash-bearing prefixes, throw or otherwise return a clear unsupported-provider error instead of silently falling through to HybridAI.
- Avoid rebuilding provider dispatch structures on every resolution call.

## Acceptance Criteria

- `resolveProviderForModel` is O(1) for prefixed provider ids.
- Unknown `provider/model` prefixes do not resolve to HybridAI.
- Bare model names still resolve to HybridAI by convention.
- Enabled/disabled provider behavior remains intact for optional providers.
- Tests cover:
  - known prefixed provider dispatch,
  - typoed/unknown prefixed provider rejection,
  - bare HybridAI passthrough,
  - local model discovery fallback if preserved.

## Notes

Estimated implementation time: ~1 hour.

furukama:
## Summary

- Replaces linear provider matching in `src/providers/factory.ts` with module-level provider and prefix registries.
- Resolves active prefixed model ids through a direct prefix map while preserving bare HybridAI passthrough behavior.
- Rejects unknown slash-bearing provider prefixes consistently through `resolveProviderForModel`, `resolveModelProvider`, `modelRequiresChatbotId`, and runtime credential resolution.
- Preserves the previous compatibility behavior for known-but-disabled provider prefixes by falling back to HybridAI.
- Keeps discovered bare local models routable via `getLocalModelInfo`.

Fixes #1112.

## Validation

Local Node 22 checks run in `/Users/bkoehler/src/hybridclaw-issue-1112`:

- `npm run typecheck`
- `npm run lint`
- `npm run check`
- `./node_modules/.bin/vitest run tests/providers.factory.test.ts tests/local-provider.test.ts tests/model-catalog.test.ts tests/gateway-status.test.ts`

GitHub Actions on `bf997ce4`:

- `lint`
- `test`
- `docker-preflight (agent, ./container, ./container/Dockerfile, runtime, hybridclaw-agent)`
- `docker-preflight (gateway, ., ./Dockerfile, hybridclaw-gateway)`
- `npm-e2e`

### DIFF
--- a/src/providers/factory.ts
+++ b/src/providers/factory.ts
@@ -1,26 +1,9 @@
-import {
-  DASHSCOPE_ENABLED,
-  DEEPSEEK_ENABLED,
-  GEMINI_ENABLED,
-  HUGGINGFACE_ENABLED,
-  HYBRIDAI_ENABLE_RAG,
-  HYBRIDAI_MODEL,
-  KILO_ENABLED,
-  KIMI_ENABLED,
-  LOCAL_LLAMACPP_ENABLED,
-  LOCAL_LMSTUDIO_ENABLED,
-  LOCAL_OLLAMA_ENABLED,
-  LOCAL_VLLM_ENABLED,
-  MINIMAX_ENABLED,
-  MISTRAL_ENABLED,
-  OPENROUTER_ENABLED,
-  XAI_ENABLED,
-  XIAOMI_ENABLED,
-  ZAI_ENABLED,
-} from '../config/config.js';
+import { HYBRIDAI_ENABLE_RAG, HYBRIDAI_MODEL } from '../config/config.js';
+import { getRuntimeConfig } from '../config/runtime-config.js';
 import { anthropicProvider } from './anthropic.js';
 import { huggingfaceProvider } from './huggingface.js';
 import { hybridAIProvider } from './hybridai.js';
+import { getLocalModelInfo } from './local-discovery.js';
 import { ollamaProvider } from './local-ollama.js';
 import {
   llamacppProvider,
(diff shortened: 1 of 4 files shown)


furukama:

furukama:
Addressed the Claude review feedback in `e79f8e19`:

- extracted shared model normalization,
- simplified prefix map keys to bare provider IDs and removed slash stripping,
- added an explicit comment for tolerant read-only provider classification,
- made unexpected discovered local backends fail fast instead of falling back to HybridAI.

There were no inline review threads to resolve; Claude posted the feedback as a top-level PR comment.

PR #1140, opened 2026-05-30
feat: suggest update on interactive startup

maxnoller:
## What

Prompt the user to update HybridClaw when a newer release is available and they start it **interactively** — `hybridclaw tui`, `hybridclaw gateway`, or `hybridclaw gateway start`. Answer `y` to install, rebuild native deps, run the postinstall, and exit (relaunch on the new version); `n` to continue on the current version.

## How it works (cache-first, mirroring Codex)

Modeled on `codex-rs/tui/src/updates.rs`:

- The prompt is decided from a **cached** latest version at `<runtime-home>/version-check.json` (`{ latestVersion, lastCheckedAt }`, **20h TTL**).
- When the cache is missing or stale, a **detached** `__refresh-version-cache` child process (`detached: true`, `stdio: 'ignore'`, unref'd) refreshes it via `npm view` for the *next* launch.
- **Startup never blocks on the network.** First run with no cache stays silent and warms the cache; the banner appears next launch.
- The actual install reuses the existing `hybridclaw update` machinery (package-manager detection, native rebuild, postinstall, gateway restart).

## Gating

- Only on an interactive terminal (both stdin and stdout TTY).
- Only for **global package installs** — source checkouts and unknown installs are skipped (no prompt, no refresh).
- Non-interactive shells / the detached gateway daemon never prompt.

## Differences from Codex (intentional, not implemented here)

- Binary `[y/N]` prompt rather than the 3-option modal with "skip this version".
- No config opt-out flag yet.
- npm registry as the only version source (Codex tailors per install method).
- Also covers the gateway entry points, not just the TUI.

## Tests

- `tests/update.test.ts`: 15 tests — non-TTY, source checkout (no prompt **and** no refresh), missing-cache→spawn refresh, stale→refresh, fresh→no refresh, up-to-date→silent, newer→prompt→decline, and `refreshVersionCache` write / no-write-on-failure. Original 5 `runUpdateCommand` tests intact.
- `tests/cli.test.ts`: mock updated for the new exports.

## Validation

- `tsc --noEmit`: clean
- `tsc --noEmit --noUnusedLocals --noUnusedParameters`: clean
- `biome check` (changed files): clean
- `vitest tests/cli.test.ts tests/update.test.ts`: **168/168 pass**

Not run: full `npm run test:unit` suite, `npm run build`.

### DIFF
--- a/docs/content/reference/commands.md
+++ b/docs/content/reference/commands.md
@@ -517,6 +517,16 @@ running with a replayable launch command, HybridClaw restarts it automatically
 with the original parameters; otherwise it falls back to manual restart
 instructions. Source checkouts receive git-based update instructions instead.
 
+When you start HybridClaw interactively (`hybridclaw tui`, `hybridclaw gateway`,
+or `hybridclaw gateway start`) and a newer release is available, HybridClaw
+prints the available version and prompts you to update before continuing. Answer
+`y` to install and restart, or `n` to start on the current version. This prompt
+appears only on an interactive terminal and only for global package installs;
+non-interactive shells and source checkouts are never blocked. The registry
+check is cached under the runtime home (`version-check.json`, 20-hour TTL) and
+refreshed in a detached background process, so startup never waits on the
+network — a newly published version surfaces on the next launch.
+
 ## Network Policy
 
 ```bash
--- a/docs/content/reference/commands.md
+++ b/docs/content/reference/commands.md
@@ -520,9 +520,13 @@ instructions. Source checkouts receive git-based update instructions instead.
 When you start HybridClaw interactively (`hybridclaw tui`, `hybridclaw gateway`,
 or `hybridclaw gateway start`) and a newer release is available, HybridClaw
 prints the available version and prompts you to update before continuing. Answer
-`y` to install and restart, or `n` to start on the current version. This prompt
-appears only on an interactive terminal and only for global package installs;
-non-interactive shells and source checkouts are never blocked. The registry
+`y` to install the update, or `n` to start on the current version. Installing
+exits the current command once the update finishes (the running process still
+holds the old code), so relaunch HybridClaw to use the new version; an
+already-running gateway is restarted automatically with its original
+parameters. This prompt appears only on an interactive terminal and only for
+global package installs; non-interactive shells and source checkouts are never
+blocked. The registry
 check is cached under the runtime home (`version-check.json`, 20-hour TTL) and
 refreshed in a detached background process, so startup never waits on the
 network — a newly published version surfaces on the next launch.
--- a/src/update.ts
+++ b/src/update.ts
@@ -101,7 +101,16 @@ function findNearestPackageRoot(startPath: string | undefined): string | null {
 
   let current: string;
   try {
-    const resolved = path.resolve(startPath);
+    // Resolve symlinks first: a global install puts a bin shim at
+    // `process.argv[1]` (e.g. /usr/local/bin/hybridclaw) whose directory has no
+    // package.json above it. Following the link lands on the real
+    // node_modules/.../dist entry so the package root is actually found.
+    let resolved = path.resolve(startPath);
+    try {
+      resolved = fs.realpathSync(resolved);
+    } catch {
+      // The path may not exist yet (or in tests); fall back to the literal path.
+    }
     current =
       fs.existsSync(resolved) && fs.statSync(resolved).isDirectory()
         ? resolved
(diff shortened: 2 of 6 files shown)


src/update.ts
```diff
@@ -460,6 +464,145 @@ function runUpdateInstall(command: UpdateCommand): Promise<number> {
...
+): Promise<void> {
+  // Only suggest updates in an interactive terminal. Non-TTY launches
+  // (CI, the detached gateway daemon, scripts) must never block on a prompt.
+  if (!process.stdin.isTTY || !process.stdout.isTTY) return;
+
+  const packageName = resolvePackageName(process.argv[1]);
+  const install = detectInstallContext(packageName, process.argv[1]);
+  // Only global package installs can be updated via the package manager.
+  // Source checkouts update via git; unknown installs are left untouched.
+  if (install.kind !== 'package') return;
```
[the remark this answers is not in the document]
maxnoller: Fixed in 546f0ad8. `findNearestPackageRoot()` now `realpath`s the entry before walking, so a global bin shim (e.g. `/usr/local/bin/hybridclaw`) resolves to the real `node_modules/.../dist/cli.js` and the package root is found — `detectInstallContext()` returns `package` instead of `unknown`. Added a test that points `process.argv[1]` at a bin symlink and confirms the prompt appears.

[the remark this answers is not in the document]
maxnoller: Fixed in 546f0ad8. The startup path now confirms the entry point is the package manager's *global* install before prompting: `isGlobalPackageInstall()` compares the resolved install root against `npm root -g` (via the existing `resolveUpdatedPackageRoot`) and bails out otherwise, so a project-local dependency or npx cache no longer triggers a global `npm install -g`. The `npm root -g` lookup runs lazily — only once the cache says an update is available — to keep it off the common startup path. Added a test for a project-local copy whose global root differs.

docs/content/reference/commands.md
```diff
@@ -517,6 +517,16 @@ running with a replayable launch command, HybridClaw restarts it automatically
 with the original parameters; otherwise it falls back to manual restart
 instructions. Source checkouts receive git-based update instructions instead.
 
+When you start HybridClaw interactively (`hybridclaw tui`, `hybridclaw gateway`,
+or `hybridclaw gateway start`) and a newer release is available, HybridClaw
+prints the available version and prompts you to update before continuing. Answer
+`y` to install and restart, or `n` to start on the current version. This prompt
```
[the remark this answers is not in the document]
maxnoller: Fixed in 546f0ad8. Reworded the docs: answering `y` installs the update and then *exits* the current command (the running process still holds the old code), so you relaunch HybridClaw to use the new version — it does not relaunch the TUI or start the gateway. An already-running gateway is restarted automatically.

src/update.ts
```diff
@@ -99,7 +101,16 @@ function findNearestPackageRoot(startPath: string | undefined): string | null {
 
   let current: string;
   try {
-    const resolved = path.resolve(startPath);
+    // Resolve symlinks first: a global install puts a bin shim at
```
furukama: We should not use inline comments. Rather have fuction names that are crystal clear.

src/update.ts
```diff
@@ -99,7 +101,16 @@ function findNearestPackageRoot(startPath: string | undefined): string | null {
...
+    // `process.argv[1]` (e.g. /usr/local/bin/hybridclaw) whose directory has no
+    // package.json above it. Following the link lands on the real
+    // node_modules/.../dist entry so the package root is actually found.
+    let resolved = path.resolve(startPath);
+    try {
+      resolved = fs.realpathSync(resolved);
+    } catch {
+      // The path may not exist yet (or in tests); fall back to the literal path.
```
furukama: Catch path should at least log some debug stuff

src/update.ts
```diff
@@ -460,6 +498,155 @@ function runUpdateInstall(command: UpdateCommand): Promise<number> {
...
+    // --inspect would make the child fail to bind its debug port and exit
+    // before it can refresh the cache.
+    const child = spawn(
+      process.execPath,
+      [cliEntry, REFRESH_VERSION_CACHE_COMMAND],
+      { detached: true, stdio: 'ignore' },
+    );
+    child.unref();
+  } catch {
+    // A failed background refresh must never affect startup.
```
furukama: Should debug log failure

furukama (changes requested):

src/update.ts
maxnoller: Fixed in 9d9808e7. Pulled the symlink resolution out of `findNearestPackageRoot` into a dedicated `resolveEntryPathThroughBinSymlink(entryPath)` helper so the name carries the intent — the explanatory inline comment is gone.

src/update.ts
maxnoller: Fixed in 9d9808e7. The `realpathSync` failure now `logger.debug`s (entry path + error) before falling back to the literal path, instead of swallowing it silently.

src/update.ts
maxnoller: Fixed in 9d9808e7. The background-refresh spawn failure now `logger.debug`s the error. Did the same for the sibling `writeVersionCache` catch so both silent failure paths log consistently.

PR #1141, opened 2026-05-31
fix: censor /secret value in web chat page

maxnoller:
## Summary

Running `/secret set <name> <value>` from the web console (the chat page) left the **plaintext secret visible** in two places:

- **The output** — the echoed user message bubble showed `/secret set NAME my-real-secret` verbatim (and the copy/replay actions carried it too).
- **The chat history page** — if a secret command ever reached persisted history, it was re-rendered verbatim on reload.

This PR masks the secret value everywhere it is shown on the chat page, while still sending the **real** value to the gateway so the secret is stored exactly as before.

Example: `/secret set TS_AUTHKEY tskey-abc123` now renders as `/secret set TS_AUTHKEY ••••••`.

## How

- New pure helper `console/src/lib/redact-secret-command.ts` → `redactSecretCommand(text)`.
  - Only `secret set` carries a plaintext value, so only its value is masked. `secret list` / `show` / `unset` / `route` reference a secret **by name** (no value) and are returned unchanged, as is any non-command message.
  - Operates line-by-line, case-insensitive, and masks the whole value (handles spaces/quotes).
- `use-chat-stream.ts`: build the echoed bubble, `rawContent`, and `replayRequest` from the masked copy; keep sending the real `content` in the `/api/chat` request body.
- `chat-history-query.ts`: mask user content when hydrating reloaded history (defense-in-depth).

## Notes

- Web slash commands are handled by the gateway and are **not** persisted to the messages table, so the live echo is the primary leak surface; the history-hydration masking is a belt-and-suspenders guard.
- Scope is intentionally limited to `/secret set` per the request. `/env set` (documented as *plaintext* runtime values) is left as-is.

## Testing

- New unit tests for `redactSecretCommand` (9 cases: slash/no-slash, quoted/multi-token values, case-insensitivity, multi-line, non-value subcommands, incomplete command, ordinary messages).
- New `useChatStream` integration test: the echoed bubble is masked **and** the wire payload still contains the real value.
- `npx vitest run` (console): **648 passed**; `tsc --noEmit` clean; `biome check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Fixed
+
+- **Censor `/secret set` in the web chat**: The value of a `/secret set <name>
+  <value>` command is now masked in the echoed message bubble and in reloaded
+  chat history, so the plaintext secret no longer lingers on the chat page. The
+  real value is still sent to the gateway so the secret is stored as before.
+
 ## [0.21.0](https://github.com/HybridAIOne/hybridclaw/tree/v0.21.0) - 2026-05-29
 
 ### Added
--- a/console/src/lib/redact-secret-command.test.ts
+++ b/console/src/lib/redact-secret-command.test.ts
@@ -37,10 +37,23 @@ describe('redactSecretCommand', () => {
     );
   });
 
-  it('only masks the secret-set line in a multi-line message', () => {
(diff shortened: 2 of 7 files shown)


furukama:
@maxnoller Slash commands should not be included in the history at all because they're ephemeral. They also should not be displayed at all on reload. Can you check why this isn't the case?

PR #1187, opened 2026-06-09
[codex] Forward response ratings to HybridAI feedback

furukama:
## Summary

- Forward accepted response ratings to the HybridAI `/api/chat_feedback` endpoint when HybridAI auth and a bot id are active, regardless of which model produced the response.
- Include the stored previous user message and rated assistant response so HybridAI receives the same prompt/response pair used by the existing feedback API.
- Prefix forwarded `bot_response` values with the rated response agent, for example `[main] ...`, when an agent id is available.
- Resolve the feedback bot id from the rated session first, then the configured observability/HybridAI bot id fallbacks; clears and missing prerequisites stay local only.

## Validation

- `mise exec node@22 -- npx vitest run tests/response-ratings.test.ts`
- `mise exec node@22 -- npm run typecheck`
- `mise exec node@22 -- npm run lint`
- `mise exec node@22 -- npx biome check src/gateway/response-ratings.ts tests/response-ratings.test.ts`

### DIFF
--- a/src/gateway/response-ratings.ts
+++ b/src/gateway/response-ratings.ts
@@ -1,9 +1,11 @@
 import { makeAuditRunId, recordAuditEvent } from '../audit/audit-events.js';
 import {
   clearResponseRating,
+  getResponseRatingFeedbackContext,
   getResponseRatingTarget,
   upsertResponseRating,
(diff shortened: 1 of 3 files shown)


furukama:

PR #1197, opened 2026-06-10
fix: harden one-line installer and install/setup paths

maxnoller:
## Summary

Deep review (`/code-review xhigh`) of the curl|bash installer and the surrounding install machinery, with every confirmed finding fixed. 15 findings survived adversarial verification; the standouts:

### Installer (`scripts/install.sh`)
- **Drop the consumer npm ≥11.10 gate and forced `npm install -g npm@^11`** — contradicted #1153 (consumer npm pin deliberately removed; the published shrinkwrap pins the tree), mutated users' global npm, and added a needless hard-failure path.
- **`CI=false` / `NO_PROMPT=0` no longer force headless mode** — headless detection is now value-aware like the sibling flags.
- **`HYBRIDCLAW_NODE_VERSION` is now honored as an exact pin** (previously only an offline fallback, despite the `--help` text). Latest-22.x discovery now uses one ~4KB `latest-v22.x/SHASUMS256.txt` fetch reused for checksum verification, replacing the ~320KB `index.json` regex scrape + second request. Offline fallback bumped 22.20.0 → 22.22.3 (two security releases behind).
- **`tar --no-same-owner`** — root installs previously left `~/.hybridclaw/node` owned by the tarball's uid 1000 (local-tamper risk on multi-user hosts).
- **Managed-Node smoke test** — a failed `$(node --version)` inside the `ok` line didn't trip `set -e`, so old-glibc hosts got "✓ Node.js  ready" and a misleading npm error later.
- **Stale-binary shadowing fixed** — a `hybridclaw` from a previous install elsewhere on PATH used to suppress PATH persistence and get verified/onboarded as if it were the new install.
- **Ctrl-C during onboarding no longer kills the installer** (`trap : INT`); the advertised "skip and continue" now works (empirically verified — bash kills the parent when a child dies of SIGINT).
- **`/dev/tty` probed by opening it** — `-r` passes without a controlling terminal where the redirect then fails with ENXIO.
- **nvm detection was dead code** (`have nvm` can't see a shell function); now probes `NVM_DIR`/`~/.nvm/nvm.sh`.
- **PATH persistence covers fish and macOS bash** (`config.fish` via `fish_add_path`; `~/.bash_profile`, which shadows `.profile`/`.bashrc` for login shells).
- npm prefix fallback now discloses the permanent `~/.npmrc` write, the undo command, and the nvm incompatibility; unset `HOME` gets a clear error; Alpine hint notes 3.21+; armv7l warns about unavailable features; `--no-audit --no-fund`; misc dedup.

### Postinstall (`scripts/postinstall-container.mjs`)
- **Dependency probe was wrong in both directions**: `require.resolve('dompurify/package.json')` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` (exports map), so dompurify was *permanently missing* → the bootstrap re-ran on every install and failed hard in airgapped setups; meanwhile the `paths` option walks up to the outer package's `node_modules`, so genuinely missing deps could count as *present* (version-skew risk). Now checks `container/node_modules/<dep>/package.json` directly.
- **yarn installs no longer break**: npm detection is an allowlist (npm user-agent / npm-cli execpath) instead of a pnpm blocklist, so yarn/bun fall back to real npm rather than running `yarn.js` with npm-only flags (reproduced: yarn 1.22 exits 1 → whole install fails).

### Homebrew (`packaging/homebrew/hybridclaw.rb`)
The HEAD-only formula could never work: `std_npm_args` packs with `--ignore-scripts`, so the gitignored `dist/cli.js` was never built; the keg-only `node@22` was never wired into the bin stubs; and the test block asserted `/hybridclaw/i` against a bare-semver `--version`. Rewritten to install the published registry tarball (0.23.0, sha256-pinned, prebuilt `dist/`), wrap bins via `write_env_script`, and assert the version.

### Runtime + docs
- `Reinstall HybridClaw.` dead-end (which reproduces the problem for `--ignore-scripts`/pnpm installs) replaced with the shipped repair command: `node "$(npm root -g)/@hybridaione/hybridclaw/scripts/postinstall-container.mjs"`. Docs gained the matching caveat.
- `scripts/check-node-version.mjs` now derives the required major from `engines.node` (same source of truth as `src/node-version-guard.ts`) and uses `process.exitCode` so piped stderr isn't truncated.

### Tests
- Offline unit tests now pin the `.tar.gz` download contract and `REQUIRED_NODE_MAJOR` ↔ `engines.node` consistency (both previously guarded only by the Docker e2e suite), plus pin/parse/fallback cases for `resolve_node_version`; stale `.tar.xz` fixture fixed.
- e2e: `docker run --init` so timed-out installs actually tear down (`--rm` never fired for PID-1 bash ignoring SIGTERM); gateway stdout drained (latent 64KB-pipe deadlock); `npm pack` destination quoted; npm-upgrade assertions replaced; yarn fallback case added.

## Deliberately not done (needs its own PR)
- Process-scoped npm prefix instead of the permanent `~/.npmrc` write — `hybridclaw update` resolves the prefix from npm config at update time, so this needs a coordinated `src/update.ts` change.
- Auto self-heal of container deps at gateway start (the repair script is idempotent and ships in the tarball; messages/docs now point at it).
- e2e download caching (derived Docker image + npm cache volume; ~6-7 min/run is mostly re-downloads).

## Test plan
- [x] `bash scripts/install.test.sh` — 25/25 pass
- [x] `vitest --project unit` for postinstall-container, host-runtime-setup, host/container-runner redaction, cli host-runtime tests
- [x] `bash scripts/install.sh --help` and `--dry-run --no-prompt` — clean, zero writes
- [x] Live `resolve_node_version`: latest 22.22.3 from one 41-line fetch; `HYBRIDCLAW_NODE_VERSION` pin honored
- [ ] `npm run test:install-e2e` (Docker matrix: ubuntu no-xz, node:22 prefix fallback, alpine musl, dry-run)
- [ ] `brew install` from the updated formula once the tap exists


maxnoller:
## Test-suite review (second pass, same `xhigh` machinery)

Reviewed the install/setup test suite itself in `ca5471d1`. Highlights:

**Defects in the first commit's own changes (caught and fixed):**
- The container-deps repair hint used `npm root -g` — wrong for exactly the pnpm installs the message names. Runtime messages now print the actual install-root script path; docs note the pnpm variant.
- The prefix-fallback e2e test title still advertised the removed npm self-upgrade.

**Verified coverage gap:** reverting the postinstall probe back to `require.resolve` passed **every existing test** — the false-present fix had zero regression coverage. Added a test (a dep present only in the outer package's `node_modules` must count as missing) and mutation-verified it fails against the revert.

**Silent-green CI:** `describe.skipIf(!dockerAvailable())` + `--passWithNoTests` meant a Docker outage (or a glob rename) turns the only job running the installer matrix into a permanent silent pass. The suite now fails loudly in CI when Docker is missing, and the flag is gone.

**Installer unit tests: 25 → 58 assertions**, covering the previously untested `NO_PROMPT`/`is_truthy` derivation (`CI=false` stays interactive), argument parsing, the `NODE_SHASUMS` checksum-cache paths, the actual nodejs.org URLs requested (all prior stubs ignored curl's arguments), `add_to_path` (idempotency, login-rc creation, `.bash_profile`, fish), `ensure_node`, and `npm_global_bin` caching.

**e2e robustness:** retry on `EINTEGRITY`/`ERR_SOCKET_TIMEOUT`/503/`fetch failed`/timeout kills; apt/apk stderr kept; containers named `hc-e2e-install-*` and swept; rc-file persistence and Alpine no-partial-install now asserted; SIGKILL/rmSync race fixed in the npm e2e; `spawnSync` timeout in the node-guard integration test; `container/package.json` required by release-check.

**Skipped (noted, not fixed):** the pnpm e2e case still asserts nothing about container deps (right contract under pnpm's script-blocking is debatable); `npm pack` mutating repo `dist/` via prepack is inherent to packing from the repo; `getAvailablePort` TOCTOU is documented as accepted in the helper; base-image tags left unpinned.

PR #1215, opened 2026-06-15
[codex] Fix hatching chat history sync

furukama:
## Summary

Fixes web chat sync issues found while testing new-agent hatching and agent creation:

- Force a fresh history fetch after `/agent switch` so gateway autostart hatching messages appear without requiring a browser reload.
- Avoid rewriting ordinary user messages like `Hi` as `@test1 Hi` when the assistant response came from the active session agent.
- Show gateway-authored hatching startup state immediately, including the model-generated prelude.
- Keep the hatching spinner visible for the whole `starting` phase, including after the prelude arrives while the full model turn is still generating.
- Make follow-up hatching turns continue from chat history instead of reintroducing the agent or repeating questions already asked.
- Keep `/agent switch` status text from mentioning hatching when no startup bootstrap file remains.
- Make onboarding emails choose a tailored subject instead of reusing a fixed default.
- Invalidate the chat agent selector after successful `/agent create` or `/agent install` commands so new agents appear without reloading `/chat`.
- Accept `old_text`/`new_text` aliases in the container `edit` tool so model-generated file edits do not fail with bogus "Text not found" errors.

## Root Cause

The chat page reused cached history via React Query after an agent switch. Because chat history is cached with `staleTime: Infinity`, the UI could miss backend-stored bootstrap messages until a full reload forced `/api/history`.

Separately, the reload history mapper inferred that a user message addressed an agent whenever the following assistant message had an `agent_id`. In switched sessions, that `agent_id` is just the active session agent, so plain messages were incorrectly decorated with `@agent`.

The hatching spinner was tied to `starting && !preludeMessageId`, so it disappeared as soon as the quick prelude was stored even though the full bootstrap turn was still running.

For follow-up messages while `BOOTSTRAP.md` still exists, the injected hatching prompt told the model to introduce itself and begin onboarding again. The session history was present, but the wrapper overrode the intended continuation behavior.

Typed `/agent create` commands completed through the stream path, which only refreshed recent chats. The `['agents-list', token]` selector query was left cached, so the new agent did not appear until reload or stale-time expiry.

Finally, in `sess_20260615_193526_4ec40116`, the model called `edit` with `old_text`/`new_text` arguments. The tool schema historically used `old`/`new`; because extra args were accepted but ignored, the tool searched for an undefined replacement target and reported "Text not found" even though the requested text existed.

## Validation

- `npm run format`
- `npm run typecheck`
- `npx vitest run tests/container.edit-tool.test.ts`
- `npm --workspace console run test -- src/routes/chat/use-chat-stream.test.ts`
- `npm --workspace console run test -- src/routes/chat/chat-page.test.tsx`
- `npx vitest run tests/gateway-service.context-references.test.ts`
- `npx vitest run tests/workspace-bootstrap.test.ts`
- `npx vitest run tests/gateway-service.agent-command.test.ts tests/gateway-service.bootstrap-autostart.test.ts`
- `npm --workspace console run test -- src/routes/chat/chat-history-query.test.ts src/routes/chat/message-block.test.tsx`
- `git diff --check`


furukama:

PR #1216, opened 2026-06-15
[codex] Recover stale Codex auth during model discovery

furukama:
## Summary

Fix Codex model discovery when HybridClaw's stored Codex access token is unexpired locally but rejected by ChatGPT as invalidated.

## Root Cause

HybridClaw trusted the stored JWT expiry before calling the Codex `/models` endpoint. In practice, ChatGPT can invalidate the access token and refresh token before the JWT expiry, causing model discovery to fail with 401/403 while local auth status still reported healthy credentials.

## Changes

- Add a forced Codex credential refresh path for stale-but-unexpired tokens.
- Retry Codex model discovery once after a 401/403 using refreshed credentials.
- Clear unusable HybridClaw Codex credentials when refresh requires relogin.
- Fall back to importing the local Codex CLI auth store non-interactively when available.
- Keep auth-failure discovery fallbacks uncached so later recovery is not masked.
- Add focused auth/discovery tests for refresh, relogin, CLI import fallback, and uncached fallback behavior.

## Validation

- `vitest tests/codex-auth.test.ts tests/codex-discovery.test.ts tests/provider-discovery-utils.test.ts`
- `tsc --noEmit`
- `npm run lint`
- Live local verification: imported valid Codex CLI credentials and confirmed Codex `/models` returned HTTP 200 without printing secrets.

### DIFF
--- a/src/auth/codex-auth.ts
+++ b/src/auth/codex-auth.ts
@@ -730,20 +730,48 @@ function needsRefresh(credentials: CodexStoredCredentials): boolean {
   return credentials.expiresAt <= Date.now() + CODEX_REFRESH_SKEW_MS;
 }
 
-export async function ensureFreshCredentials(): Promise<EnsureFreshCredentialsResult> {
-  const initial = assertStoredCredentials(loadCodexAuthStore());
-  if (!needsRefresh(initial)) {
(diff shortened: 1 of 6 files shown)


furukama:

PR #1217, opened 2026-06-15
[codex] Fix legacy audit user actor indexing

furukama:
## Summary

- Canonicalize legacy structured-audit `userId` fields into local user actors for actor indexing.
- Preserve the signed audit payload shape instead of injecting a new `actor` field.
- Add regression coverage for live writes and v43 migration/backfill using the web user ID shape that triggered the warning.

## Root Cause

Structured audit ingestion treated legacy plain `userId` values like `web-user-b903d512` as canonical user IDs. Actor validation requires `username@authority`, so live `session.start` writes logged `Structured audit event has invalid actor fields` and skipped actor indexing.

## Validation

- `npm run format`
- `npx vitest run --configLoader runner tests/structured-audit-query.test.ts`
- `git diff --check`

Note: `npm run format` completed successfully and reported existing optional-chain warnings elsewhere in the repo; it did not modify files.

### DIFF
--- a/src/memory/db.ts
+++ b/src/memory/db.ts
@@ -10480,14 +10480,28 @@ function readPayloadActor(
   }
 
   try {
-    return actorFromLegacyFields({
-      userId: readPayloadStringValue(payload, 'userId'),
-      agentId: readPayloadStringValue(payload, 'agentId'),
-    });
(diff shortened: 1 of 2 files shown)


furukama:

PR #1221, opened 2026-06-16
[codex] Add logs page logging mode controls

furukama:
## Summary

Adds a logging mode control to the admin Logs page so operators can switch logging Off, On, or Debug from the console. The control saves runtime config and reloads the gateway so the change takes effect without restarting from the CLI, and the log viewer stays pinned to the newest tail content when logs load or refresh.

## Details

- Adds persisted `ops.logRequests` and `ops.debugModelResponses` flags alongside `ops.logLevel`.
- Wires request logging and model-response debug capture to runtime config while preserving the existing environment/CLI overrides.
- Maps Logs page modes to config values: Off -> silent/no capture, On -> info/no capture, Debug -> debug plus request/model response capture.
- Keeps mode mapping in one typed table and does not treat a manually configured `trace` log level as the Debug mode.
- Treats CLI `--debug` as a startup log level instead of a permanent forced override, so `/admin/logs` can lower the level after a foreground debug start.
- Resyncs the logger from runtime config during admin reload, including the case where the persisted config value did not change.
- Derives the displayed mode from effective runtime logging state so true forced runtime settings are shown as Debug instead of On.
- Auto-scrolls the selected log tail to the end after load or refresh.
- Keeps tailed log content line-aligned when the tail starts mid-file, avoiding partial first lines like a clipped `DEBUG` token.
- Prints full local dates in formatted log timestamps instead of time-of-day only.
- Strips ANSI escape sequences from admin log tails and disables color output when stdout is redirected to `gateway.log`.
- Delays CLI logger construction so foreground gateway runs can attach the configured gateway log file before the logger stream is created.
- Centralizes model-response debug enablement and adds invalid-env-value warnings.
- Checks the gateway reload response before showing the logging update as saved.
- Narrows segmented toggle callbacks so route-level runtime guards and casts are not needed.
- Caches the admin config query briefly while still refreshing logs after save to read the effective post-reload runtime state.
- Updates `config.example.json` and adds backend and console coverage.

## Validation

- `npx vitest run tests/config-reload.integration.test.ts tests/gateway-service.audit.test.ts tests/model-response-debug.test.ts`
- `../node_modules/.bin/vitest run src/routes/logs.test.tsx`
- `./node_modules/.bin/vitest run tests/gateway-http-server.test.ts -t "returns admin log file metadata and selected tail"`
- `./node_modules/.bin/vitest run tests/logger-format.test.ts tests/logger.test.ts`
- `./node_modules/.bin/vitest run tests/logger.test.ts tests/model-response-debug.test.ts tests/gateway-http-server.test.ts -t "returns admin log file metadata and selected tail|reloads gateway config|logger|model response debug"`
- `./node_modules/.bin/tsc --noEmit -p console/tsconfig.json`

## Notes

A Vite browser smoke test reached the admin auth/connect shell, but without a live gateway API behind the local frontend there was no meaningful Logs page render to inspect.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -581,7 +581,9 @@
     "gatewayBaseUrl": "http://127.0.0.1:9090",
     "gatewayApiToken": "",
     "dbPath": "~/.hybridclaw/data/hybridclaw.db",
-    "logLevel": "info"
+    "logLevel": "info",
+    "logRequests": false,
+    "debugModelResponses": false
   },
   "observability": {
     "enabled": true,
--- a/console/src/api/client.ts
+++ b/console/src/api/client.ts
@@ -23,6 +23,7 @@ import type {
   AdminChannelTransport,
   AdminCommandResult,
   AdminConfig,
+  AdminConfigReloadResponse,
   AdminConfigResponse,
   AdminCreateSkillPayload,
   AdminDistillConsentPayload,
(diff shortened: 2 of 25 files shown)


furukama:

PR #1231, opened 2026-06-17
[codex] Fix mobile chat QR session auth

furukama:
## Summary

Fix the mobile chat QR take-over flow so cloud/container deployments preserve Gateway web authentication after scanning the QR code.

## Root Cause

The `/chat/continue?token=...` endpoint only wrote the target chat user/session into `localStorage`. On cloud instances, `/chat/<session>` is protected by Gateway session auth, so the browser reached the right chat URL without a valid `hybridclaw_session` cookie and was redirected into the HybridAI login/API-key flow instead of opening the chat.

## Changes

- Store the current web-session auth payload with each one-time mobile launch token.
- Set a signed HttpOnly `hybridclaw_session` cookie when redeeming the QR token before redirecting to `/chat/<session>`.
- Preserve the existing one-time token and expiry behavior.
- Add a Docker/container-mode regression test that redeems the QR code and verifies `/chat/<session>` loads with the issued session cookie.

## Validation

- `npx vitest run tests/gateway-http-server.test.ts --testNamePattern mobile`
- `npx vitest run tests/gateway-http-server.test.ts`
- `npm run format`
- `git diff --check`

### DIFF
--- a/src/gateway/auth-token.ts
+++ b/src/gateway/auth-token.ts
@@ -30,6 +30,10 @@ function readSharedSecret(): string {
   return (process.env.HYBRIDCLAW_AUTH_SECRET || '').trim();
 }
 
+export function hasSharedAuthSecret(): boolean {
+  return Boolean(readSharedSecret());
+}
+
 function parseSignedToken(token: string): ParsedToken | null {
   const trimmed = token.trim();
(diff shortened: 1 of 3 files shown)


furukama:

PR #1253, opened 2026-06-22
[codex] Advertise public A2A Agent Card URL

furukama:
## Summary

- Use the configured `deployment.public_url` when rendering `/.well-known/agent.json`.
- Keep request/proxy-header origin reconstruction as the fallback for local/dev cases.
- Add a regression test for sandbox requests where the incoming `Host` is an internal container address but the configured public URL is HTTPS.

## Why

Sandbox instances were publishing Agent Cards with internal delivery URLs like `http://172.19.x.x:9090/a2a`. The A2A trust flow correctly rejects those cards because delivery URLs must use HTTPS unless targeting loopback, so pairing from `/admin/a2a-trust` failed even when the operator entered the public HTTPS sandbox URL.

## Validation

- `./node_modules/.bin/vitest tests/gateway-http-server.test.ts --run`

### DIFF
--- a/src/gateway/gateway-http-server.ts
+++ b/src/gateway/gateway-http-server.ts
@@ -2298,11 +2298,18 @@ function resolveRequestOrigin(
   return `${proto}://${host}`;
 }
 
-function resolveA2AAgentCardOrigin(req: IncomingMessage): string {
-  return (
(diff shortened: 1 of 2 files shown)


furukama:

PR #1263, opened 2026-06-23
[codex] Provision A2A delegation trust during pairing

furukama:
## Summary

- Advertise the public A2A delegation-token verifier key and sender agent ids on HybridClaw Agent Cards.
- Carry that delegation trust advertisement through operator pairing requests.
- Provision inbound trusted A2A sender records when a peer is trusted or an incoming pairing request is approved.
- Update A2A pairing and cross-instance outbound integration tests so delivery no longer depends on manual trust-ledger seeding.

## Root Cause

The A2A Trust UI only populated the peer instance public-key ledger. Inbound `/a2a` delivery verifies signed delegation JWTs against the separate trusted A2A sender ledger, so a recipient could show the sender instance as trusted while rejecting message delivery with HTTP 401.

## Validation

- `npx vitest run tests/a2a-pairing.test.ts tests/a2a-outbound.integration.test.ts`
- `npm run lint`
- `git diff --check`


furukama:

PR #1267, opened 2026-06-25
[codex] Make allowed senders editable in admin

furukama:
## Summary

- Replaces raw sender allowlist textareas in the admin channel settings with an explicit add/remove list editor.
- Applies the editor to DM/group sender allowlists across chat, email, Slack, Teams, and iMessage channel settings while leaving non-sender list fields as textareas.
- Updates the channel settings test to cover adding multiple Signal allowed senders through the UI.

## Why

The underlying config already stores allowed senders as arrays, but the admin UI made adding more senders implicit by requiring comma/newline editing in a textarea. This makes the affordance explicit without changing the saved config shape or runtime authorization behavior.

## Impact

Admins can add and remove multiple allowed senders directly in `/admin/channels`. Existing configs remain compatible because the API payload still uses the same `string[]` fields.

## Validation

- `npx biome check --write console/src/routes/channels.tsx console/src/routes/channels.test.tsx console/src/styles.css`
- `npm --workspace console run test -- src/routes/channels.test.tsx`

### DIFF
--- a/console/src/routes/channels.test.tsx
+++ b/console/src/routes/channels.test.tsx
@@ -1098,7 +1098,7 @@ describe('ChannelsPage', () => {
           daemonUrl: 'http://127.0.0.1:8080',
           account: '+14155550123',
           dmPolicy: 'allowlist',
-          allowFrom: ['+14155551212'],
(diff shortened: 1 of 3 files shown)


furukama:

PR #1275, opened 2026-06-27
fix(mcp): prefer GATEWAY_PUBLIC_BASE_URL for OAuth redirect URI

maxnoller:
## Problem

When the gateway runs behind a reverse proxy, the MCP OAuth `redirect_uri` is derived from the incoming request's `Host` (via `resolveRequestOrigin` → `X-Forwarded-Host` → `Host`). If the proxy doesn't forward `X-Forwarded-Host`, the `Host` the gateway sees is its **internal** listen address, so it builds an unreachable redirect:

```
http://172.19.0.22:9090/api/mcp/oauth/callback
```

The user's browser (reaching the gateway via its public address) can't route there, so the `?code=` redirect after consent dies and OAuth-based MCP connects (Sentry's hosted MCP, Linear, Notion, …) can't complete.

`resolveMcpOAuthRedirectUri` already had a `GATEWAY_BASE_URL` fallback, but it never helped here: `requestBaseUrl` is always non-empty (it falls back to `127.0.0.1`), so the configured value was never reached.

## Fix

Add a `GATEWAY_PUBLIC_BASE_URL` env var, applied in `config.ts` alongside `HEALTH_HOST`. When set, `resolveMcpOAuthRedirectUri` **prefers** it over the request-derived origin:

```ts
String(GATEWAY_PUBLIC_BASE_URL || requestBaseUrl || GATEWAY_BASE_URL || '')
```

The redirect URI is computed once at authorize time and persisted on the flow record (`mcp-oauth.ts`), then reused verbatim at token exchange — so authorize and exchange stay consistent.

This is **defense-in-depth**: OAuth correctness doesn't depend solely on the upstream proxy forwarding `X-Forwarded-Host`. When `GATEWAY_PUBLIC_BASE_URL` is unset, behavior is unchanged (derive from the request), so standalone/desktop runs are unaffected.

## Tests

`tests/mcp-oauth-redirect.test.ts`:
- request origin used when no public base URL is configured (current behavior preserved),
- configured public base URL preferred over an internal request origin,
- trailing slash trimmed.

`tsc --noEmit` (incl. `--noUnusedLocals --noUnusedParameters`) clean; existing `tests/mcp-oauth.test.ts` still green.

## Docs

Documented the env var in `docs/content/developer-guide/runtime.md`.

### DIFF
--- a/docs/content/developer-guide/runtime.md
+++ b/docs/content/developer-guide/runtime.md
@@ -111,6 +111,14 @@ Core details:
   already-running HybridClaw gateway.
 - `HEALTH_HOST` can override the health server bind address without rewriting
   runtime config, which is useful in containerized or proxied deployments.
+- `GATEWAY_PUBLIC_BASE_URL` is the public origin the gateway is reached at when
+  it runs behind an ingress proxy (e.g. `https://u-<id>.sbx.example.com`). When
+  set it is preferred over the request-derived origin for building
+  externally-reachable URLs — notably MCP OAuth redirect URIs, which otherwise
+  fall back to the incoming request's `Host`/`X-Forwarded-Host`. Set this when a
+  proxy can't be relied on to forward `X-Forwarded-Host`, so OAuth callbacks
+  don't end up pointing at the gateway's internal address. Leave it unset for
+  standalone/desktop runs.
 - `mcpServers.*.env` and `mcpServers.*.headers` are persisted in
   `~/.hybridclaw/config.json` exactly as configured today. Treat them as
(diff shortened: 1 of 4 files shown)


maxnoller:
Closing — superseded by a fix at the ingress-proxy layer that forwards `X-Forwarded-Host`, which lets the gateway build the correct OAuth redirect URI from the request with no extra config. That resolves the issue for proxied deployments, so this `GATEWAY_PUBLIC_BASE_URL` change is redundant there.

It still has standalone value for a gateway sitting behind a proxy that can't forward that header, so reopen if that need arises.

PR #1276, opened 2026-06-27
[codex] Use public URL for MCP OAuth callbacks

furukama:
## What changed

- Prefer `deployment.public_url` when building MCP OAuth callback URLs, so cloud admin flows do not hand providers an internal container address.
- Fall back through public gateway/request origins before private/local origins.
- Teach request-origin resolution to prefer the configured public deployment URL for other externally consumed admin URLs.
- Add targeted gateway and OAuth redirect tests for cloud/public URL behavior.

## Why

Cloud admin requests can arrive at the gateway with an internal `Host` such as `172.19.0.21:9090`. The MCP OAuth start flow used that request origin first, so providers like Zoho redirected back to an internal address that a browser cannot reach.

## Impact

Cloud deployments with `deployment.public_url` set now generate public callback and launch URLs for external OAuth/browser flows. Local development still falls back to the configured/request base URL when no public deployment URL is present.

## Validation

- `biome check` on the touched gateway and test files
- `vitest run tests/gateway-service.mcp-oauth-redirect.test.ts`
- `vitest run tests/gateway-http-server.test.ts -t "configured public URL|mobile chat QR uses configured public URL"`
- `tsc --noEmit`

### DIFF
--- a/src/gateway/gateway-http-server.ts
+++ b/src/gateway/gateway-http-server.ts
@@ -2295,6 +2295,11 @@ function resolveRequestOrigin(
   const explicitBaseUrl = normalizePublicBaseUrl(bodyBaseUrl);
   if (explicitBaseUrl) return explicitBaseUrl;
 
+  const configuredPublicUrl = normalizePublicBaseUrl(
+    getRuntimeConfig().deployment.public_url,
+  );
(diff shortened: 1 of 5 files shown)


furukama:

PR #1287, opened 2026-07-03
[codex] fix stalled Codex stream timeout

furukama:
## Summary

Fixes a silent hang in the OpenAI Codex streaming provider. The Codex SSE reader now uses the same idle-timeout wrapper as the other streaming providers, so a stream that emits an initial event and then stalls surfaces a retryable timeout instead of leaving the gateway turn open indefinitely.

## Root Cause

`container/src/providers/openai-codex.ts` called `reader.read()` directly. If the upstream SSE connection emitted `response.created` and then stopped sending bytes without closing, the model call could remain open for several minutes with no visible user progress or final error.

## Validation

- `npx vitest run tests/openai-codex-provider.test.ts`
- `npm run lint`
- `npm run format`

### DIFF
--- a/console/src/api/apps.ts
+++ b/console/src/api/apps.ts
@@ -1,0 +1,96 @@
+import { requestJson } from './client';
+
+export type AppCategory =
+  | 'apps'
+  | 'documents'
+  | 'games'
+  | 'productivity'
+  | 'creative'
+  | 'quiz'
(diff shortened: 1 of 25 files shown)


furukama:
Superseded by #1288. This PR was accidentally opened from an existing feature-branch base and included unrelated app-gallery changes.

PR #1291, opened 2026-07-06
[codex] Preserve full tool progress output

furukama:
## What changed

- Emit complete tool-call arguments and tool results in container progress logs using a line-safe `json:<JSON string>` payload.
- Decode line-safe progress payloads in the gateway parser before streaming and persisting activity traces.
- Add regression coverage for multiline JSON tool-result previews so pretty-printed results no longer collapse to a dangling `{`.

## Why

The activity trace UI was receiving only the first physical stderr line for pretty-printed tool results. For JSON results, that meant the persisted preview could be just `{` even though the full result was available in audit logs.

## Impact

Web/chat activity traces now preserve full tool calls and results instead of truncated previews. Existing legacy progress lines still parse unchanged.

## Validation

- `npx vitest run tests/tool-progress-parser.test.ts`
- `npm run build`
- `npm run format`
- `npm run lint`
- `git diff --check`

Note: `npm run format` completed with existing optional-chain warnings and no fixes applied.

### DIFF
--- a/container/src/index.ts
+++ b/container/src/index.ts
@@ -574,26 +574,21 @@ interface CompletedToolCallExecution {
   artifacts: ArtifactMetadata[];
 }
 
-function logToolCallStart(
-  toolName: string,
-  argsJson: string,
-  approval: ToolApprovalEvaluation,
-): void {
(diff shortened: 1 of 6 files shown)


furukama:

PR #1292, opened 2026-07-06
[codex] Fix email polling with stale mailbox UID state

furukama:
## Summary

- Refresh IMAP mailbox UID metadata with `STATUS` before each email poll.
- Search only the UID range after the persisted cursor instead of trusting cached `mailbox.uidNext`.
- Add a regression test for a new message being missed when the selected mailbox UID state is stale.

## Root Cause

The email poller could skip `SEARCH` when `client.mailbox.uidNext` had not been refreshed on a long-lived IMAP connection. Pressing "Save channel settings" restarted the email integration, which refreshed the mailbox state and caused the pending reply to be processed.

## Validation

- `npm run format`
- `npm run lint`
- `/Users/bkoehler/src/hybridclaw/node_modules/.bin/vitest --root /Users/bkoehler/.codex/worktrees/d2f4/hybridclaw --config /Users/bkoehler/src/hybridclaw/vitest.config.ts run tests/email.connection.test.ts`

### DIFF
--- a/src/channels/email/connection.ts
+++ b/src/channels/email/connection.ts
@@ -65,19 +65,36 @@ function resolveCursorStatePath(address: string): string {
   );
 }
 
-function resolveMailboxUidNext(client: ImapFlow): number {
-  return Math.max(1, client.mailbox ? client.mailbox.uidNext : 1);
+function normalizeUidNext(value: unknown): number {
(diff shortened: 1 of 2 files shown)


furukama:

PR #1295, opened 2026-07-06
Fix non-final assistant drafts leaking into chat

furukama:
## Summary

Fixes a chat-boundary bug where non-final assistant draft text could be surfaced as the assistant bubble when the tool loop continued or stalled.

## Root Cause

This was not a real final message with a STOP marker. The container loop cached stripped assistant text before deciding whether the turn was a Ralph STOP, a continuation draft, or a tool-request turn. The stalled/overflow fallback then reused that cached draft as `output.result`, and streamed text deltas could cross the gateway boundary before final classification.

## Changes

- Adds explicit assistant segment classification: `draft`, `tool_request`, or `final`.
- Adds `outputPresentation` metadata to container output, gateway chat results, and streamed text events so rendering is driven by metadata rather than output-string heuristics.
- Suppresses provider text deltas until the model turn is classified as final, then emits only the sanitized final segment text with `assistant_bubble` metadata.
- Tracks `latestFinalAssistantText` separately and uses it for overflow/stalled fallback output.
- Updates the console stream client to pass text-event metadata through and hide metadata-hidden outputs without parsing message text.
- Removes the previous Codex-specific tool-markup parser approach from this PR.
- Adds Ralph boundary and metadata tests for STOP/final, draft/no-STOP, tool-request, approval, and assistant-bubble presentation.

## Validation

- `hybridclaw gateway status`
- `./node_modules/.bin/vitest run tests/container.ralph.test.ts tests/openai-codex-provider.test.ts tests/tool-call-normalizer.test.ts tests/gateway-client.test.ts`
- `npm --workspace console test -- src/lib/chat-stream.test.ts src/routes/chat/use-chat-stream.test.ts`
- `npm run format`
- `npm run lint`
- `npm run check`

Note: `npm run format` / `npm run check` still report existing optional-chain suggestions and exit 0; no unrelated optional-chain rewrites were included.

### DIFF
--- a/console/src/api/chat-types.ts
+++ b/console/src/api/chat-types.ts
@@ -135,9 +135,24 @@ export interface ChatArtifact {
   type?: string;
 }
 
+export type OutputSegmentKind =
+  | 'draft'
+  | 'final'
+  | 'tool_request'
+  | 'approval'
+  | 'status';
+export type OutputDisplaySurface = 'none' | 'assistant_bubble' | 'approval';
+
+export interface OutputPresentationMetadata {
+  segmentKind: OutputSegmentKind;
+  visible: boolean;
+  displaySurface: OutputDisplaySurface;
+}
+
 export interface ChatStreamTextDelta {
   type: 'text';
   delta: string;
+  outputPresentation?: OutputPresentationMetadata;
 }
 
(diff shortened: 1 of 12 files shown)


furukama:

PR #1298, opened 2026-07-06
[codex] Dispatch inbound A2A inbox messages

### LINKED ISSUE #716: Agent inbox delivery semantics and dispatcher — push vs. pull
## Context

#425 (PR #695) ships in-process A2A primitives: `sendMessage(envelope)` and
`inbox(agentId)`. The primitives are correct, but the question of *when* an
agent sees inbox messages during a running turn is unanswered.

Current implementation also accepts signed inbound A2A/webhook envelopes and
persists them to the local A2A thread store, but there is no receiver-side
dispatcher that consumes those persisted messages and starts or resumes work for
the intended `recipient_agent_id`. Inbound delivery currently means "saved and
visible in inbox/admin surfaces," not "the target agent is now working on it."

## Decision needed

When agent B receives a message from agent A, how does agent B see it?

- **Option A — pull.** The agent must explicitly call an `inbox` tool to read
  pending messages. Simple, explicit, agent-controlled. Risk: messages sit
  unread because the agent has no reason to poll.
- **Option B — push.** The harness injects pending inbox messages into the
  agent's next prompt before each turn. No-op when inbox is empty. Risk:
  prompt bloat, ordering with user input, inbox-loop hazards (replying to a
  message itself triggers a reply turn).
- **Option C — hybrid.** Push a *summary* (count + senders + intents) into
  the system/prefix; agent reads full content via `inbox` tool. Compromise on
  prompt size.

We also need to decide the trigger/dispatcher model:

- **No automatic dispatch:** A2A messages are inbox/card artifacts only until a
  human, schedule, or agent turn pulls them.
- **Automatic dispatch:** inbound A2A creates a durable work item and a
  dispatcher invokes the recipient agent.
- **Board-mediated dispatch:** inbound A2A creates/updates an R29 triage card;
  card claim/assignment semantics decide when the recipient agent runs.

## Receiver dispatcher scope

If we choose automatic or board-mediated dispatch, implement a receiver-side
dispatcher that:

- maps `recipient_agent_id` to a local registered agent and fails closed for
  unknown or non-local recipients;
- creates or resumes the right session/thread for the A2A conversation;
- calls the normal agent execution path for the recipient agent with the A2A
  sender, thread id, intent, and content available in the prompt/context;
- records durable processed/read state so retransmits or retries are
  idempotent;
- audits dispatch start, success, failure, and any ignored/duplicate envelope;
- handles retry/backoff and operator-visible failure state;
- enforces loop protection so agent-to-agent replies do not create unbounded
  ping-pong;
- has intent-aware behavior for `chat`, `handoff`, `escalate`, and `ack`;
- is covered by tests for inbound envelope -> persisted message -> dispatched
  recipient run, duplicate delivery, unknown recipient, failed run retry/failure
  state, and loop suppression.

## Sub-questions

- Does a turn start *because* of an inbox message, or only on human /
  scheduled trigger? (The trigger model is the harder half.)
- Once an envelope is read, is it marked read? `inbox()` currently returns
  *all* envelopes ever received — needs a `since` / `after_id` cursor or a
  read marker to support push/dispatch mode.
- What is the durable ownership model: A2A thread owner, recipient agent inbox,
  R29 card owner, or active session agent?
- Intent-aware routing: should `escalate` produce a different UX than `chat`?
- Interaction with R29 Jobs surface (#687): when an operator can see a message
  before the recipient agent does, what is the source of truth?

## Why this matters

This decides what an "agent team" actually feels like in HybridClaw. It is
the user-facing slice of R1 — Principle IV.

Without the dispatcher decision, A2A can transport and persist messages, but it
cannot yet guarantee that the addressed agent actually starts work.

## Depends on / blocks

- Depends on #425 (primitives).
- Blocks #713 (agent tool needs a delivery model to wire into).
- Blocks any meaningful R29 (#687) treatment of A2A messages on the kanban
  surface.
- Blocks treating inbound A2A as fully functional delegated work rather than
  stored coordination messages.

furukama:
## Summary

- add a durable A2A inbox dispatch queue for persisted local deliveries
- dispatch pending inbound A2A messages to the addressed local agent through the normal gateway message path
- add idempotency, retry/backoff, terminal failure/ignored/suppressed states, audit events, and thread loop suppression
- cover dispatch, duplicate delivery, unknown recipient, failed recipient run, ignored ack, and loop suppression behavior

## Validation

- `npx vitest run tests/a2a-inbox-dispatcher.test.ts --reporter=dot`
- `npx vitest run tests/a2a-inbox-dispatcher.test.ts tests/a2a-runtime.integration.test.ts tests/a2a-inbound.test.ts tests/gateway-service.a2a-inbox.test.ts --reporter=dot`
- `npm run typecheck`
- `npm run lint`

Closes #716

### DIFF
--- a/console/src/api/chat-types.ts
+++ b/console/src/api/chat-types.ts
@@ -193,6 +193,19 @@ export type ChatStreamEvent =
 
 export type ChatResultMessageRole = 'assistant' | 'approval' | 'command';
 
+export type A2ADeliveryState = 'pending' | 'delivered' | 'failed' | 'unknown';
+
+/**
+ * Describes an outbound A2A message a chat send produced, so the UI can render
+ * a live delivery-status chip and poll for its final state.
+ */
+export interface A2ADeliveryDescriptor {
+  messageId: string;
+  threadId: string;
+  recipientAgentId: string;
+  status: A2ADeliveryState;
+}
+
 export interface ChatStreamResult {
   status?: string;
   error?: string;
@@ -214,6 +227,7 @@ export interface ChatStreamResult {
   artifacts?: ChatArtifact[];
   apps?: Array<{ id: string; title: string; kind: 'web' | 'live' }>;
   toolsUsed?: string[];
+  a2aDelivery?: A2ADeliveryDescriptor | null;
 }
 
 export interface MediaItem {
@@ -260,4 +274,5 @@ export interface ChatMessage {
   addressedAgentPresentation?: AssistantPresentation | null;
   responseRating?: ResponseRatingValue | null;
   branchKey?: string | null;
+  a2aDelivery?: A2ADeliveryDescriptor | null;
 }
--- a/console/src/api/client.ts
+++ b/console/src/api/client.ts
@@ -1,4 +1,5 @@
 import type {
+  A2ADeliveryStatusResponse,
   AdminA2AInboxResponse,
   AdminA2APairingPreviewResponse,
   AdminA2APairingStartRequest,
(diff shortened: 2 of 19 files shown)


furukama:

PR #1337, opened 2026-07-18
feat: implement Phase 2 deterministic model routing

furukama:
## Summary

- add the bundled deterministic `tier-router` middleware and automatically enable it with model routing
- route heartbeat, scheduler, full-auto, and default auxiliary tasks from the bottom tier
- use agent models as start-tier preferences while preserving request, session, and onboarding pins as hard overrides
- execute agent turns through within-tier provider fallback and across-tier escalation
- add sticky tiers, de-escalation, and the one-shot `/escalate` command
- record `route.escalated` audit events and per-attempt routing fields in model usage telemetry

## User-facing impact

Yes, but there are no console or UI changes in this phase:

- operators can use `/escalate` to start the next unpinned agent turn one tier higher
- routed turns can transparently retry weak empty/narrate-only responses and escalate on classified auth, rate-limit, 5xx, or malformed-tool failures
- an agent's configured model is now a routing start preference rather than a hard pin; explicit request and session model selections remain pinned
- route tier, zone, reason, and escalation status are observable in audit and usage records

## Safety and failure boundaries

- never retry after a tool execution or while an approval is pending
- buffer streamed attempt events and discard deltas from failed attempts
- do not escalate unknown provider errors
- fail closed if routing is enabled while the bundled tier router is disabled or replaced
- keep explicit request/session model pins outside the ladder
- bound in-memory sticky and pending-escalation state

## Validation

- `npm run format`
- `npm run lint`
- `npm run build`
- targeted routing, gateway, fallback, task taxonomy, middleware, and concierge tests: 111 passed
- console tests: 830 passed
- desktop tests: 25 passed
- `git diff --check`

A full `npm run test:unit` run encountered four failures in `tests/gateway-service.fullauto.test.ts`. All four reproduced with every Phase 2 tracked implementation file restored to `HEAD`, so they are existing baseline failures rather than changes introduced here.

The live-fleet golden E2E exit gate was not run locally because it requires the deployed model fleet and credentials.

### DIFF
--- a/container/shared/provider-fallback.d.ts
+++ b/container/shared/provider-fallback.d.ts
@@ -1,4 +1,4 @@
-export type FallbackReason = 'auth' | 'rate_limit' | 'other';
+export type FallbackReason = 'auth' | 'rate_limit' | 'server_error' | 'other';
 
 export function classifyProviderError(err: unknown): FallbackReason;
 export function shouldFallbackProviderError(err: unknown): boolean;
--- a/container/shared/provider-fallback.js
+++ b/container/shared/provider-fallback.js
@@ -1,12 +1,24 @@
 export function classifyProviderError(err) {
   const text = err instanceof Error ? err.message : String(err);
   if (/(^|\D)401(\D|$)|(^|\D)403(\D|$)/.test(text)) return 'auth';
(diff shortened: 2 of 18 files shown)


furukama:

PR #1347, opened 2026-07-18
Stabilize conversation prefixes for Anthropic prompt caching

furukama:
## Summary

- move per-turn time, summaries, retrieved context, and daily memory after retained conversation history, next to the latest user turn
- truncate history by dropping complete oldest turns without rewriting retained message bytes
- keep the skills catalog stable on explicit skill-invocation turns
- split Anthropic system prompts into volatility-ordered static core, workspace/memory, and skills cache blocks
- add the fourth standard cache breakpoint to the last stable message block, including tool-loop results
- preserve existing collapsed-system behavior for non-Anthropic providers and document the prompt-cache contract

## Why

Dynamic context previously appeared immediately after the system message, so every turn changed bytes near the start of the prompt. History optimization also sliced individual messages, continuously changing retained prefixes. Anthropic received one large system block and no multi-turn message breakpoint, so a memory update invalidated the entire system cache and tool-loop/follow-up growth produced cache writes instead of reads.

## Impact

Retained conversation history is append-only until complete old turns are evicted. Anthropic can reuse the static system prefix across workspace-memory changes, reuse earlier system boundaries across skills changes, and reuse stable conversation prefixes across tool iterations and follow-up turns.

## Validation

- `npm run lint`
- `npm run build`
- 454 focused gateway, prompt, history, and provider regression tests
- 38 final core prompt/cache regression tests after cleanup
- broad unit run exercised 5,345 tests; the affected stale assertion was fixed and rerun, and the transient embedding-provider failure passed standalone. The remaining local limitation is a host-runner fixture that requires a missing `container/node_modules/.bin/agent-browser` binary.

### DIFF
--- a/container/shared/dynamic-context.d.ts
+++ b/container/shared/dynamic-context.d.ts
@@ -1,0 +1,5 @@
+export declare const DYNAMIC_CONTEXT_MESSAGE_PREFIX = '<context>\nDate (UTC): ';
+
+export declare function isDynamicContextMessageText(
+  value: unknown,
+): value is string;
--- a/container/shared/dynamic-context.js
+++ b/container/shared/dynamic-context.js
@@ -1,0 +1,10 @@
+export const DYNAMIC_CONTEXT_MESSAGE_PREFIX = '<context>\nDate (UTC): ';
+
+export function isDynamicContextMessageText(value) {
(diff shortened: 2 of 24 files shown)


furukama:

PR #1386, opened 2026-07-27
fix(email): keep replies threaded when a send passes a bad inReplyTo

maxnoller:
## Problem

The `message` tool accepts any string as `inReplyTo`/`references` (`src/channels/message/tool-actions.ts`) and writes it straight into the outbound RFC 5322 headers. Explicit thread headers deliberately *replace* the account's tracked thread — that is how a caller replies to a message other than the latest one (`sendEmail`, `src/channels/email/delivery.ts:294-300`). So when the supplied value is not a message id, the reply is detached from the thread the runtime already knew about:

- the inbound `References` chain is dropped,
- `resolveSubjectAndBody` loses its thread context and falls back to `DEFAULT_EMAIL_SUBJECT`, so the reply goes out titled `HybridClaw` instead of `Re: <their subject>`,
- the composed context is then handed to `threadTracker.remember` (`runtime.ts:457`), so the agent's *own* following messages to that recipient inherit the break.

Nothing logs it. The agent reports a successful send, and the person who emailed the agent gets a new thread rather than an answer.

Observed in an agent mailbox — a reply to a message whose id was `<1785…@example.com>` went out as:

```
Subject: HybridClaw
In-Reply-To: <0f99ab14eeb8>
References: <0f99ab14eeb8>
```

`<0f99ab14eeb8>` is not a message id at all. The runtime had the correct inbound context in hand at send time and discarded it. It is intermittent because it depends on whether the model supplies the argument: the same flow, with `inReplyTo` omitted, threads correctly.

## Fix

Validate the ids rather than trusting them:

- **`message` send rejects** an `inReplyTo`/`references` entry that is not a message id, with an error naming where a real one comes from (`messageId` in a `message read` result) — the caller can correct it instead of silently mis-threading.
- **Delivery drops** malformed ids arriving from any other caller and falls back to the tracked thread, which is the right answer whenever the caller was answering the message that opened it.
- **A bare `local@domain` id is accepted and wrapped**, since that shape still names a real message and is a common way to write one.
- **Prompt hint and tool schema** now say to omit `inReplyTo` for replies in the current thread, and to copy a `messageId` from a read result otherwise. The old hint promised automatic threading without naming the parameter that overrides it.

Threading behaviour for well-formed explicit ids is unchanged — replying to an older message still works exactly as before, including its own subject.

## Testing

- `tests/agent-email.e2e.test.ts`: new end-to-end case — inbound mail, agent answers via the tool with `inReplyTo: <0f99ab14eeb8>`, the tool call fails with the actionable error, and the reply still goes out `In-Reply-To: <boss-runbook-1@example.com>` / `Re: Runbook request`. Shared setup extracted into helpers; the existing case is unchanged. 2/2 pass.
- `tests/email.channel.test.ts`: 3 new delivery cases — malformed explicit id falls back to the tracked thread (both headers and the remembered context), malformed ids with no tracked thread are dropped rather than emitted, bare `local@domain` is wrapped. 31/31 pass.
- Verified the new tests fail on the parent commit, reproducing the reported headers exactly: `"inReplyTo": "<0f99ab14eeb8>", "subject": "HybridClaw"`.
- `tsc --noEmit --noUnusedLocals --noUnusedParameters` and `biome check` clean; full `npm run test:unit` 5416 passed, 1 unrelated pre-existing failure (`tests/host-runner.redaction.test.ts`, agent-browser binary path, fails on the parent commit too).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@
 
 ### Fixed
 
+- **Email replies stay in their thread**: A `message` send whose `inReplyTo` or
+  `references` is not an RFC 5322 message id is now rejected with an actionable
+  error instead of being written into the outbound headers verbatim. Such a
+  value used to detach the reply from the thread the runtime was already
+  tracking — losing the `References` chain and the `Re:` subject, so mail
+  clients filed the answer as a new `HybridClaw` thread — and was then
+  remembered as the thread for every later message to that recipient. Malformed
+  ids reaching delivery from any other caller are dropped in favour of the
+  tracked thread, and a bare `local@domain` id is accepted and wrapped.
 - **Cloud admin sessions recover after inactivity**: Expired browser sessions
   reload through HybridAI sign-in instead of showing an unactionable
   `WEB_API_TOKEN` prompt. The self-hosted fallback identifies the exact gateway
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,18 +15,17 @@
 ### Fixed
 
 - **Email replies stay in their thread**: A `message` send whose `inReplyTo` or
-  `references` is not an RFC 5322 message id is now rejected with an actionable
-  error instead of being written into the outbound headers verbatim. Such a
-  value used to detach the reply from the thread the runtime was already
-  tracking — losing the `References` chain and the `Re:` subject, so mail
-  clients filed the answer as a new `HybridClaw` thread — and was then
-  remembered as the thread for every later message to that recipient. Malformed
-  ids reaching delivery from any other caller are dropped in favour of the
-  tracked thread, and a bare `local@domain` id is accepted and wrapped. The
-  `brevo-email` plugin's `send_email` validates the same way. A `message read`
-  result now reports `messageId: null` for a sent entry reconstructed from the
-  local audit log, instead of a synthetic id that names no real message and
-  cannot be replied to.
+  `references` is not an RFC 5322 message id now drops that id and threads the
+  reply on the current email thread, reporting what it dropped in the tool
+  result. Such a value used to be written into the outbound headers verbatim,
+  which detached the reply from the thread the runtime was already tracking —
+  losing the `References` chain and the `Re:` subject, so mail clients filed
+  the answer as a new `HybridClaw` thread — and was then remembered as the
+  thread for every later message to that recipient. A bare `local@domain` id
+  is accepted and wrapped. The `brevo-email` plugin's `send_email` behaves the
+  same way. A `message read` result now reports `messageId: null` for a sent
+  entry reconstructed from the local audit log, instead of a synthetic id that
+  names no real message and cannot be replied to.
 - **Cloud admin sessions recover after inactivity**: Expired browser sessions
   reload through HybridAI sign-in instead of showing an unactionable
   `WEB_API_TOKEN` prompt. The self-hosted fallback identifies the exact gateway
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,11 @@
   clients filed the answer as a new `HybridClaw` thread — and was then
   remembered as the thread for every later message to that recipient. Malformed
   ids reaching delivery from any other caller are dropped in favour of the
-  tracked thread, and a bare `local@domain` id is accepted and wrapped.
+  tracked thread, and a bare `local@domain` id is accepted and wrapped. The
+  `brevo-email` plugin's `send_email` validates the same way. A `message read`
+  result now reports `messageId: null` for a sent entry reconstructed from the
+  local audit log, instead of a synthetic id that names no real message and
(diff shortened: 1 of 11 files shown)


maxnoller:
Follow-ups from an independent review, pushed as a9560b6e.

**Fixed — the hint contradicted the validation.** `message read` could hand back a synthetic Sent-folder id (`synthetic:audit:<id>` / `synthetic:<id>`, `src/channels/email/admin-mailbox.ts:856,1147`) for entries reconstructed from the local audit log when the sent copy has not been appended to IMAP — `appendSentCopiesToImap` is best-effort and warn-only. Those ids reached the agent as `messageId` (`tool-actions.ts:863`), so an agent doing exactly what the new hint says — reuse a read result's `messageId` — would have been refused for a legitimate prior message, under a condition it cannot see. `mapEmailMailboxMessage` now reports `messageId: null` for such entries; the `folder:uid` `id` still identifies them for reads.

**Fixed — a second send path had the same bug.** The `brevo-email` plugin registers its own `send_email` tool with a separate copy of the pre-fix threading logic (`plugins/brevo-email/src/brevo-outbound.js`), so accounts on that channel were fully exposed. It now validates at the tool boundary and drops malformed ids in its transport. The earlier claim that "delivery drops malformed ids arriving from any other caller" did not hold for this caller.

**Fixed — inconsistent observability.** Malformed `references` entries were dropped silently while `inReplyTo` warned; both log now.

**Test fixes.** Renamed a delivery test that described a case it did not exercise (one valid reference survived, so headers were not omitted) and added the genuinely-headerless branch, plus a validation table covering real ids (domain-literal, IDN, bare `local@domain`) and rejected ones (internal, synthetic, session key, prose, CRLF injection).

**Not changed, with reasons:**

- *"The e2e test's mail assertion doesn't depend on the fix."* It does. `findSentMail` returns the **first** message to that recipient, which before the fix is the tool-driven send, not the runtime auto-ack. Verified by reverting `src`/`container` to the parent commit and removing only the `toolError` assertions: the run fails on the mail assertion with `subject: "HybridClaw"`.
- *Not overselling this as a security fix.* Agreed and already the case — nodemailer strips CR/LF from these headers itself, so this is a threading-correctness fix. The regex rejects CRLF values as a side effect, not as the point.
- *Bare `local@domain` widens the accept surface.* Kept deliberately: a value with exactly one `@` is far more likely a real id written without brackets than an internal token, and the alternative rejects ids that would otherwise thread correctly.

**Coverage gap, stated plainly:** the `mapEmailMailboxMessage` change has no direct test — `runEmailMailboxReadAction` needs the db + IMAP + runtime-config harness from `gateway-admin-email.test.ts`, which is disproportionate for a one-line mapping. The validator it calls is pinned by unit tests including the exact `synthetic:audit:42` shape.

Full unit suite: 5421 passed, 1 unrelated pre-existing failure (`tests/host-runner.redaction.test.ts`, fails on the parent commit too). `tsc` and `biome` clean.

maxnoller:
Reworked the failure mode in d10bf294 after a fair challenge: is hard-rejecting really better than letting the agent's own judgement work?

It isn't. Two things came out of digging into it:

**1. Refusing the send is worse than the bug it guards against.** The agent's goal is the reply. Failing the tool call because one optional argument named nothing threadable can leave the person who wrote in with no answer at all — strictly worse than the mis-threaded answer this branch started from. So a non-message-id `inReplyTo`/`references` is now dropped, the mail is sent in the tracked thread, and the drop is reported in the tool result as `ignoredThreadIds` plus a note telling the caller to omit the argument for in-thread replies or pass a `messageId` from a `message read` result. The agent learns; the user still gets their email. Wrong *types* still throw — a non-string `inReplyTo` is a schema violation, not a recoverable choice.

**2. The agent never needed the argument in the first place, and now demonstrably doesn't.** The inbound thread context is remembered per sender when mail arrives (`runtime.ts`, `threadTracker.remember(inbound.senderAddress, ...)`), and the message tool's sends go through the same module-level runtime the gateway initialises (`gateway.ts:2224`) — so once the unusable id is out of the way, a tool-driven reply threads on the inbound message by itself. The e2e test now asserts exactly that: the agent passes `<0f99ab14eeb8>`, and the mail still goes out `In-Reply-To: <boss-runbook-1@example.com>` / `Re: Runbook request`.

That second point also fixes a flaw in the test harness worth calling out: it used a private `createEmailRuntime()` instance, which gave the tool path its own empty thread tracker and made a tool-driven reply look inherently unthreadable. It now drives the module-level runtime the way the gateway does, so it exercises the real wiring.

Net effect on the prompt/schema changes from the first commit: they stay, and they now describe something true in both directions — omit the argument and threading is automatic; pass a real `messageId` and it is honoured.

Full unit suite 5421 passed (1 unrelated pre-existing failure, `tests/host-runner.redaction.test.ts`); e2e 2/2; `tsc` and `biome` clean.

PR #1422, opened 2026-08-24
fix(msteams): resolve canonical session keys in message tool sends

maxnoller:
## Problem

Agent-initiated sends into the current Teams chat via the `message` tool always failed, while the reply-pipeline consent-card delivery worked — so the agent told the user "I couldn't attach the file to this Teams chat" right before the file arrived. Observed live on 2026-08-24 in the HybridAI-tenant HybridClaw DM (and on 2026-08-21 during the #1415 file-upload verification, reported as "channel lookup failed").

## Root cause

Teams sessions have used canonical session keys (`agent:<id>:channel:msteams:chat:<type>:peer:<id>`) since the extensible session routing migration (8920fc7f), but the message-tool Teams routing still matched only the legacy `teams:` prefix:

- `container/src/tools.ts` — `TEAMS_SESSION_ID_RE = /^teams:/i` meant the sandbox tool neither advertised the current Teams conversation nor used it as the default send target.
- `src/channels/message/tool-actions.ts` — `isLikelyMSTeamsToolRequest` didn't recognize canonical session ids.
- `src/channels/msteams/tool-actions.ts` — `isMSTeamsSessionId` / `pickMostRecentMatchingSession` never matched any stored session, so resolution failed with "No known Teams conversation matched this request", and `ensureAuthorizedMSTeamsSendTarget` would have rejected the requester anyway. `channel-info` also mis-reported `isDm` for canonical keys.

Discord got the equivalent fix (`:channel:discord:` awareness) but Teams was left behind; the existing tests only used legacy `teams:dm:...` fixtures, which is why this never surfaced in CI.

## Fix

- Add shared `isMSTeamsSessionId` / `isMSTeamsDmSessionId` helpers in `src/channels/msteams/utils.ts` that accept canonical keys via `parseSessionKey` (mirroring Slack's `isSlackSessionId`) plus the legacy `teams:` prefix for pre-migration rows.
- Use them in the message-tool gate and the msteams tool actions.
- Widen the container-side regex to `/^teams:|:channel:msteams:/i` (same approach as the container's Discord check).

## Testing

- New gateway-side tests: canonical-key send with attachment, `msteams:current` targeting, and `channel-info` DM reporting.
- New container-side test: send defaults to the current Teams conversation with a canonical session key.
- `npx vitest run tests/message.tool-actions-send.test.ts tests/container.message-tool-normalization.test.ts` (68 passed) plus msteams runtime/inbound/attachments, channel-message-tool-hints, container schema, and gateway-http-server suites (452 passed); `tsc --noEmit` clean in root and container; biome clean on touched files.

The live behavior (upload direction, consent-card delivery) was verified end-to-end in Teams web today; after this lands the agent's direct send path should stop erroring and the misleading "couldn't attach" replies should disappear.

### DIFF
--- a/container/shared/msteams-session-ids.d.ts
+++ b/container/shared/msteams-session-ids.d.ts
@@ -1,0 +1,7 @@
+export declare function isMSTeamsSessionId(
+  value: string | null | undefined,
+): boolean;
+
+export declare function isMSTeamsDmSessionId(
+  value: string | null | undefined,
+): boolean;
--- a/container/shared/msteams-session-ids.js
+++ b/container/shared/msteams-session-ids.js
@@ -1,0 +1,20 @@
+import { parseSessionKey } from './session-keys.js';
+
+// Pre-migration Teams rows keep their legacy `teams:...` ids forever
+// (classifySessionKeyShape treats them as opaque), so both spellings
+// stay valid indefinitely.
+const LEGACY_MSTEAMS_SESSION_ID_RE = /^teams:/i;
+const LEGACY_MSTEAMS_DM_SESSION_ID_RE = /^teams:dm:/i;
+
+export function isMSTeamsSessionId(value) {
+  const normalized = String(value || '').trim();
+  if (LEGACY_MSTEAMS_SESSION_ID_RE.test(normalized)) return true;
+  return parseSessionKey(normalized)?.channelKind === 'msteams';
+}
+
+export function isMSTeamsDmSessionId(value) {
+  const normalized = String(value || '').trim();
+  if (LEGACY_MSTEAMS_DM_SESSION_ID_RE.test(normalized)) return true;
+  const parsed = parseSessionKey(normalized);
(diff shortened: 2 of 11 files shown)


maxnoller:
Pushed a follow-up commit (964da6e) addressing the two known compromises in the original diff:

- **One session-key implementation everywhere**: the wire-format codec (`buildSessionKey`/`parseSessionKey`) moved to `container/shared/session-keys.js` (the existing hand-written shared-module pattern, consumed by both `src/` and `container/src/`); `src/session/session-key.ts` re-exports it, so no import sites changed. The Teams predicates live in `container/shared/msteams-session-ids.js` and replace the container-side regex — the container now also recognizes positional canonical keys, which the regex missed.
- **Deduped `looksLikeMSTeamsConversationId`** into `msteams/utils.ts` (was defined identically in two tool-action files).

No gateway-side behavior change; existing session-key/message-tool/container/msteams/slack suites all pass.

furukama (approved):

PR #1448, opened 2026-09-01
feat(vonage-voice): gate inbound calls with a caller allowlist

maxnoller:
The Vonage voice channel accepts every inbound caller. Any number that reaches the linked Vonage number opens a realtime session and can consult the agent, with `maxConcurrentCalls` as the only limit. Every messaging channel already has a gate for this; voice did not.

## Change

Adds the same shape the messaging channels use (`dmPolicy` + `allowFrom`):

```yaml
callerPolicy: open | allowlist | disabled   # default: open
allowFrom: []                               # E.164 numbers, or '*'
```

Enforced in the answer webhook **ahead of the capacity check**, so a refused call does not consume a concurrency slot. The caller hears a short spoken refusal and the call ends.

Numbers are compared in a canonical `+<digits>` form, so allowlist entries written with spaces, dashes, or no leading `+` still match the value Vonage puts in `from`. A caller who withholds their number arrives with an empty `from` and therefore never matches an allowlist.

The refused number is logged. Without it the allowlist is undiagnosable — an operator has no way to learn which number to add.

Defaults to `open`, so existing deployments keep their current behaviour. Operators fronting an agent with tools will want `allowlist`.

## Tests

`tests/vonage-voice-caller-policy.test.ts` — 8 cases covering the policy matrix, formatting-tolerant matching, the wildcard, withheld numbers, normalization, and dedupe.

## Notes for reviewers

- The helpers are local to the plugin, matching the existing per-channel precedent (`normalizeAllowList` / `matchesAllowList` are defined locally in each of signal, telegram, slack, imessage rather than shared).
- An allowlist gates unknown callers; it is not a defence against caller-ID spoofing. Operators fronting anything sensitive should still restrict the agent's tools.
- Two related gaps are **not** addressed here: the admin console has no plugin-config UI at all (nothing reads `configSchema`/`configUiHints`), and `hybridclaw plugin config` cannot set array values (`plugin config.allowFrom must be array`). Today the allowlist can only be edited by hand in `plugins.list[].config`. Happy to split those out.

### DIFF
--- a/plugins/vonage-voice/hybridclaw.plugin.yaml
+++ b/plugins/vonage-voice/hybridclaw.plugin.yaml
@@ -26,6 +26,15 @@ configSchema:
     publicBaseUrl:
       type: string
       pattern: '^https://'
+    callerPolicy:
+      type: string
+      enum: [open, allowlist, disabled]
+      default: open
+    allowFrom:
+      type: array
+      default: []
+      items:
+        type: string
     mode:
       type: string
       enum: [turn, realtime]
@@ -47,6 +56,11 @@ configSchema:
 configUiHints:
   applicationId:
     label: Application ID
+  callerPolicy:
+    label: Caller Policy
+  allowFrom:
+    label: Allowed Callers
+    placeholder: '+4915123456789'
   mode:
     label: Call Mode
   fromNumber:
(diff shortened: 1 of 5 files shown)


maxnoller:
Superseded: the allowlist belongs on the core voice config rather than in this plugin. Every other channel keeps `dmPolicy`/`allowFrom` in the runtime config, Twilio voice has the same gap, and core settings already get a console editor for free via the generated settings registry. Redoing it there.

PR #1484, opened 2026-09-04
feat(gateway): persist session attachments in the workspace

maxnoller:
## Problem

After uploading a file, a follow-up such as “convert it to Excel” cannot reliably find it because transport caches live outside the agent workspace and stored chat only retains the filename.

## Behavior

Inbound local attachments are copied into `attachments/<session-hash>/<upload-hash>/<filename>` in the agent workspace. Follow-up turns receive the session directory so normal file tools can discover earlier uploads, including when another file is attached. Current-turn media points to the staged file in host or Docker mode; stored chat retains user-visible summaries.

Readable filenames are preserved, separate uploads with equal names coexist, and existing workspace files are not overwritten. The directory hint does not enumerate old PDFs, avoiding automatic extraction of historical documents on unrelated turns. No attachment-history KV index is needed.

## Retention and risk

- Workspace copies persist until explicitly deleted, independently of transport-cache expiry. Resetting a conversation selects a new directory; deleting chat history does not delete workspace files. Operators must include these copies in workspace storage, retention, and backup policies.
- Session directories organize files; they do not add an access-control boundary. Sessions sharing an agent workspace share filesystem access, and existing transport-cache mounts are unchanged.
- Staging uses existing source-path validation, rejects symlink destinations, publishes complete copies without overwriting files, and releases active requests on staging failure. Attachment names and contents remain untrusted.

## Validation

- Installed the lockfile-pinned dependencies using npm 11.10.0 in an isolated clone; no dependency artifacts changed.
- `npm run format`, `npm run lint`, and `npm run typecheck` pass. Format reports existing optional-chain warnings outside this change.
- 87 tests pass across session attachments, gateway media history, audio transcription, concierge, plugins, tier routing, and PDF context (`vitest run --configLoader runner --project unit --maxWorkers 1` with those seven suites).
- Coverage includes host/Docker path mapping, cache deletion, duplicate filenames, session/reset directory separation, lazy workspaces, traversal and symlink rejection, failed-copy cleanup, active-request cleanup, mixed current/prior attachments, and current-versus-historical PDF processing.
- Docker runtime and credential-backed live checks were not run. Docker path behavior is covered at the gateway/media boundary; runtime mounts and the IPC schema are unchanged.


furukama:
@maxnoller Maybe we don't need this anymore because the model now has awareness of its own tool calls?

maxnoller:
> @maxnoller Maybe we don't need this anymore because the model now has awareness of its own tool calls?

yeah good point, lets close this pr and if this is still an issue we can re-open it later

PR #1487, opened 2026-09-04
docs: document the OpenAI-compatible API for external callers

maxnoller:
## Why

A customer asked how an external system can send a prompt to their agent and read the response over REST. The docs could not answer it: `/v1/models` and `/v1/chat/completions` appear only in `reference/commands.md` under *Local Eval Workflows*, with a loopback example and a closing line saying they are "intended for local tooling and eval harnesses rather than public exposure".

Two things an integrator needs most were undocumented entirely:

- **how to pick which agent answers** — today that is `…__hc_eval=agent=<id>` on the model id, or the `X-HybridClaw-Eval-Profile` header (`src/evals/eval-profile.ts`, `src/gateway/openai-compatible-request.ts:848`)
- **that selecting an agent auto-approves the turn's tool calls** — `isEvalRequest` sets `autoApproveTools: true` (`src/gateway/openai-compatible.ts:307`)

## What changed

New guide `guides/openai-compatible-api.md` covering: scoped `hck_` token with the `openai.api` action, reaching the gateway over HTTPS, a minimal completion, agent selection (both forms) with the auto-approval warning, the fresh-session-per-request semantics, the delegation acknowledgement + polling path, and a troubleshooting table. Registered in the guides index and `navigation.json`.

Two corrections to `reference/commands.md`, both verified against the code rather than carried over:

- `/v1` accepts a scoped `hck_` token carrying `openai.api`, not only `WEB_API_TOKEN`/`GATEWAY_API_TOKEN` — `resolveAuthContext` + the `isAdminActionAllowed(payload, 'openai.api')` gate (`src/gateway/gateway-http-server.ts:11304`)
- with no agent selection the turn runs as the built-in `main` agent; `agents.defaultAgentId` does **not** apply here, because `openai-compatible.ts:269` falls back to the hardcoded `DEFAULT_AGENT_ID` instead of `resolveDefaultAgentId()`

The last point is documented as current behavior. Whether `/v1` *should* honour `agents.defaultAgentId` is a separate question — happy to open a follow-up if that fallback is a bug rather than a decision.

## Verification

- `npx vitest run tests/docs-viewer.test.ts` — 35 passed
- `npx vitest run tests/gateway-http-docs.integration.test.ts` — 13 passed, incl. "every markdown file in docs/content/ renders as 200 HTML" and the internal-link resolution check

Docs only; no runtime code touched.

### DIFF
--- a/docs/content/guides/README.md
+++ b/docs/content/guides/README.md
@@ -1,6 +1,6 @@
 ---
 title: Guides
-description: Practical workflows for generated apps, local providers, remote access, MCP, bundled skills, Twilio voice, local speech tooling, and optional office tooling.
+description: Practical workflows for generated apps, local providers, remote access, MCP, bundled skills, Twilio voice, local speech tooling, the OpenAI-compatible API, and optional office tooling.
 sidebar_position: 1
 ---
 
@@ -32,5 +32,7 @@ These pages focus on common operator workflows after the base install works.
 - [Voice and TTS](./voice-tts.md) for voice reply and speech backend setup
 - [Realtime Voice for Web Apps](./voice-web-api.md) for browser voice
   sessions from your own web app via scoped API tokens
+- [OpenAI-Compatible API](./openai-compatible-api.md) for sending prompts
+  to an agent from an external system over `/v1/chat/completions`
 - [Optional Office Dependencies](./office-dependencies.md) for host-side
   office tooling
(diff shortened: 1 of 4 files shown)


maxnoller:
Folded into #1488 so the guide describes the fixed default-agent behaviour rather than the old fallback.

PR #1492, opened 2026-09-07
fix(prompt): generated files must not be the only place the answer lives

maxnoller:
## Problem
When an agent answers a data question by generating a PDF/XLSX/DOCX, the final reply frequently contains only the download link ("The PDF is created: …"). In chat channels that render attachments poorly the user sees no answer at all, and evaluation graders that read the reply text cannot see the result either.

## Change
One new line in the deliverable rules of the tool-execution prompt block (`src/agent/prompt-hooks.ts`): when a generated file answers a question with data, the reply must state the key results in text before the file link. Test added next to the sibling assertions in `tests/prompt-hooks.tool-summary.test.ts`.

## Verification
- `vitest run --project unit tests/prompt-hooks.tool-summary.test.ts` — 23 passed
- `tsc --noEmit` clean


maxnoller:
Closing: this is a product preference, not a defect. The underlying eval gap is grader-side (it read only the reply text, not the generated PDF); handled in the eval harness instead, and per-agent workspace instructions cover customers who want figures restated in chat.

PR #1493, opened 2026-09-08
fix: evaluate agent-created cron tasks in the user timezone

### LINKED ISSUE #1463: Agent-created cron jobs are evaluated in UTC with no timezone parameter
## Problem

Agent-created recurring tasks are evaluated in UTC with no way to set a timezone.

- `createJob` always stores `tz: ''` (`src/memory/jobs.ts:238-256`).
- The scheduler parses DB tasks without a tz (`src/scheduler/scheduler.ts:471-473, 884-886`) and falls back to `DEFAULT_SCHEDULER_TIME_ZONE = 'UTC'` (`scheduler.ts:45`).
- Config-defined jobs support `schedule.tz`, but the `cron` tool exposes no timezone parameter (`container/src/tools.ts:5161-5170`), and the admin task API hardcodes `tz: ''` for agent tasks (`src/gateway/gateway-scheduled-task-service.ts:310-335`).
- `src/agent/prompt-hooks.ts:535` gives timezone guidance only for one-shot `at` reminders.

A user in CEST asking for a 09:00 briefing gets `0 9 * * *`, which fires at 11:00 local.

## Proposed fix

Resolve the user timezone from USER.md (the container already has `readUserTimezoneFile`) and store it in the job's `tz` when the tool creates a cron job, or add an explicit `tz` parameter to the tool. Expose `tz` for agent tasks in the admin API.

## Related

- #261 (closed) covered one-shot `at` timestamps only.
- PR #1459 adds "cron runs in UTC" to the tool description but leaves timezone-aware evaluation as a follow-up.

---

maxnoller:
## Summary

Fixes #1463. Agent-created recurring tasks were always stored with an empty timezone and evaluated in UTC, so a user in CEST asking for a 09:00 briefing got it at 11:00 local.

- The `cron` tool accepts an optional IANA `tz` argument and otherwise resolves the timezone from `USER.md` (`**Timezone:**` line). Unknown timezones are rejected before the task is queued. The result message quotes the effective timezone.
- The schedule side effect, `createJob`, `ScheduledTask`, and the runners' injected task list carry `tz`. `createJob` rejects an unknown timezone at the IPC boundary.
- The scheduler passes the task timezone to cron parsing for both next-run computation and firing, and to the fired prompt's current-time line.
- The admin scheduler API exposes the stored timezone for agent tasks, so the console shows it.
- Prompt guidance and the tool description now tell the model to write cron expressions in local time instead of converting to UTC.

Tasks without a stored timezone keep evaluating in UTC, so existing tasks are unaffected.

## Risk notes

Touches `src/gateway/` and `src/infra/` (type plumbing only) and `src/scheduler/`. The only new input is a timezone string, validated with `Intl.DateTimeFormat` in both the container and `createJob`; an invalid value fails the tool call or the side effect instead of silently falling back to UTC.

## Test plan

- [x] `npm run typecheck`, `npm run lint`, `npm --prefix container run lint`, `npm run build`
- [x] `vitest run tests/container.cron-tool.test.ts tests/agent-side-effects.test.ts tests/scheduler.cron.test.ts tests/memory-jobs.test.ts tests/gateway-service.scheduler.test.ts tests/scheduler.timestamp.test.ts` (32 passed)
- New tests: explicit `tz`, unknown `tz`, USER.md default, task list output, side-effect persistence and rejection, next-run computation in `Europe/Berlin`


maxnoller:
Folded into #1490: the timezone handling now flows through the synchronous `POST /api/scheduler/task` path instead of the removed schedule side-effect, so keeping this as a separate PR would only create conflicts. #1490 closes #1463.

PR #1497, opened 2026-09-08
fix(memory): serialize concurrent workspace memory writes

### LINKED ISSUE #1477: Per-agent workspace memory files are written concurrently by chat, heartbeat and scheduler with non-atomic read-modify-write
## Problem

Serialization is per session (`gatewaySessionQueue.run(req.sessionId, …)`, `src/gateway/gateway-chat-service.ts:763,772`; `containerSessionQueue`, `src/infra/container-runner.ts:178`), but the workspace is per agent (`container-runner.ts:653-661` → `src/infra/ipc.ts:360-370`). Scheduled runs use isolated sessions (`cron:<taskId>`, `src/gateway/gateway.ts:3932-3936`) and heartbeat has its own, so chat turns, heartbeat runs and N scheduled tasks for one agent execute concurrently in the same directory.

Writes are non-atomic read-modify-write on both sides: `memory append`/`replace` in the container (`container/src/tools.ts:3016-3028, 3054-3062`) and host-side consolidation (`src/memory/memory-consolidation.ts:684-693`, `readFileSync` then `writeFileSync` with no lock or rename). Concurrent appends can lose entries.

## Proposed fix

Per-agent write lock for workspace memory files, or write-to-temp plus atomic rename, on both the container and host paths.

---

maxnoller:
Concurrent chat, heartbeat, and scheduled sessions share an agent workspace. Memory updates now take a cross-process file lock across read/modify/write and publish with atomic rename, preventing lost appends and partial reads. Model consolidation rejects a rewrite when MEMORY.md changed during its model call.

Closes #1477.

Validation: 26 targeted tests passed (memory tool, consolidation, separate-process contention, failed rename, and IPC); root lint/typecheck, container lint, and full build passed. Live provider and Docker deployment tests were not needed for the filesystem transaction change.

Risk notes: locks coordinate the memory tool and consolidation, not arbitrary shell writes or external editors. Busy tools retry for five seconds; consolidation skips contention. Locks left by crashed writers require manual removal after verifying the writer has stopped; no age-based lock stealing. Memory path restrictions and approval policy remain unchanged.

### DIFF
--- a/container/shared/memory-file.d.ts
+++ b/container/shared/memory-file.d.ts
@@ -1,0 +1,3 @@
+export function lockMemoryFile(filePath: string): () => void;
+export function waitForMemoryFileLock(filePath: string): Promise<() => void>;
+export function writeMemoryFileAtomic(filePath: string, content: string): void;
--- a/container/shared/memory-file.js
+++ b/container/shared/memory-file.js
@@ -1,0 +1,57 @@
+/**
+ * Cooperative memory-file transactions shared by host and container processes.
+ * Locks cover the entire read/modify/write; rename alone cannot prevent lost updates.
+ * This is not a workspace access policy and cannot serialize external editors.
+ */
+import { randomUUID } from 'node:crypto';
+import fs from 'node:fs';
+import path from 'node:path';
+import { setTimeout } from 'node:timers/promises';
+
+export function lockMemoryFile(filePath) {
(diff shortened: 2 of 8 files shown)


maxnoller:
Folded into #1499, which now targets main and carries the whole memory stack.

PR #1498, opened 2026-09-08
fix(memory): make daily note saves safe and timezone consistent

### LINKED ISSUE #1467: memory tool: write overwrites the daily note, default target MEMORY.md is always rejected, container/host timezone fallback mismatch
## Problem

Three ergonomics gaps in the container `memory` tool and its date handling (`container/src/tools.ts`):

1. `memory write` replaces the entire daily note with `fs.writeFileSync` and reports "Wrote N chars" (`:3032-3042`). The tool schema lists `write` as an ordinary action with no warning (`:4070-4073`). A model that means "save this" wipes the day's earlier notes.
2. The default target when no `file_path`/`target` is given is `MEMORY.md` (`:2129, 2136-2137`), but every write action rejects MEMORY.md (`:2955-2962`). The natural call `memory append` without a path always errors and costs a wasted tool round-trip.
3. Daily-note file naming depends on the `**Timezone:**` line in USER.md on both sides (`:2060-2098`, `src/workspace.ts:990-996`). When it is empty, the container falls back to its process zone (UTC, no TZ passed via `src/infra/container-runner.ts:760-808`) and the host to its local zone, so around midnight the tool appends to one date file while the prompt loader and consolidation read another. The dynamic context header also always says `Date (UTC):` (`container/shared/dynamic-context.js:1`) regardless of the user timezone, so the model can compute the wrong filename near midnight.

## Proposed fix

- Make `write` require an explicit confirmation flag, or drop it in favour of `append`/`replace`.
- Default write actions to today's daily note.
- Pass the resolved user timezone (or `TZ`) into the container and use it host-side as the fallback; print the date in the user timezone in the context header, or state the daily-note filename explicitly.

---

maxnoller:
Saving a note without a target now appends to today's daily file, while untargeted reads still use MEMORY.md. Whole-file `write` requires `confirm_overwrite: true`, and the schema explains that it erases earlier entries. Invalid explicit paths fail rather than silently falling back.

Docker inherits the host's resolved timezone via TZ, with USER.md still taking precedence. Blank timezone fields cannot consume the next line. Dynamic context names the daily file explicitly across UTC midnight.

Closes #1467. Stacked on #1497; merge that PR first.

Validation: 27 targeted tests passed, covering writes, overwrite confirmation, invalid paths, blank/invalid timezones, midnight context, Docker environment, prompt stability, and IPC. Root lint/typecheck, container lint, and full build passed. Live providers and a real Docker deployment were not exercised.

Risk notes: today's-note path restrictions and approval tiers remain intact. The inherited timezone affects newly launched containers; existing containers retain their launch environment. No gateway restart is performed by this change.

### DIFF
--- a/container/shared/workspace-time.js
+++ b/container/shared/workspace-time.js
@@ -71,7 +71,7 @@ function getTimezoneOffsetMs(timezone, date) {
 
 export function extractUserTimezone(content) {
   if (typeof content !== 'string' || !content.trim()) return undefined;
-  const match = content.match(/\*\*Timezone:\*\*\s*(.+)/i);
+  const match = content.match(/\*\*Timezone:\*\*[ \t]*([^\r\n]*)/i);
   const timezone = match?.[1]?.trim();
   return timezone || undefined;
 }
--- a/container/src/tools.ts
+++ b/container/src/tools.ts
@@ -2220,6 +2220,7 @@ function resolveMemoryFilePath(args: Record<string, unknown>): string | null {
     normalizeMemoryFilePath(args.file_path) ||
     normalizeMemoryFilePath(args.path);
   if (direct) return direct;
+  if (args.file_path !== undefined || args.path !== undefined) return null;
 
   const target =
     typeof args.target === 'string' ? args.target.trim().toLowerCase() : '';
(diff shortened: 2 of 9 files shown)


maxnoller:
Folded into #1499, which now targets main and carries the whole memory stack.

maxnoller:
Folded into #1499, which now targets main and carries the whole memory stack.

PR #1501, opened 2026-09-08
feat(ratings): forward the rater's display name to HybridAI chat feedback

maxnoller:
## Why

Ratings from chat channels (Teams `/thumbs` and 👍/👎 reactions) reach the HybridAI feedback board with the rater's opaque directory id as `external_user_id`. The board cannot resolve that to a person, so every rating looks like it came from the agent owner.

## What

- `submitResponseRating` / `applyReactionRatingChanges` accept an optional `operatorDisplayName`.
- The `/thumbs` command passes the request username; the Teams reaction handler passes the reactor's username.
- The chat feedback payload gains `external_user_name` when a display name is present and differs from the user id (channels fall back to the raw id as username, which would add nothing).
- Corrected the stale comment on `resolveHybridAIChatFeedbackBrowserId`: DM session keys embed the peer id, so it was never free of user identifiers.
- Teams channel docs mention that forwarded feedback carries the rater's id and name.

## Tests

- new case: display name forwarded, id-equal name dropped
- existing payload assertions unchanged (no name → no field)

### DIFF
--- a/docs/content/channels/msteams.md
+++ b/docs/content/channels/msteams.md
@@ -182,7 +182,9 @@ Two ways to rate a HybridClaw answer from Teams:
 
 Ratings are stored per user and feed the same response-rating pipeline as the
 web console thumbs (audit trail, adaptive-skill feedback, and HybridAI chat
-feedback forwarding).
+feedback forwarding). The forwarded feedback carries the rater's Teams user id
+and display name so the HybridAI feedback board shows who rated an answer
+rather than the agent owner.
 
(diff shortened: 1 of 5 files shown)


maxnoller:
Counterpart on the platform side accepts and displays the field.

maxnoller:
Closing: forwarding the rater's display name would move from a pseudonymous id to directly identifying personal data by default for every Teams deployment. The platform side now shows the external rater id instead (snoller/chat#2317), which is what was asked for. Revisit as an explicit opt-in if a customer needs names.

PR #1502, opened 2026-09-08
fix(voice): stop realtime responses playing over each other during consults

maxnoller:
## Problem

On a phone call in realtime mode, while a `consult_agent` call is outstanding, two model responses could play at the same time: the out-of-band "still working" reassurance and a server-VAD auto-response triggered by any caller noise or backchannel ("mhm", "ja"). Their audio deltas were interleaved into the same playback queue, which the caller hears as garbled mumbling. Both responses were also free to invent a result the tool had not returned yet.

`response.cancel` without a `response_id` only stops the conversation's own response, so barge-in never reached an out-of-band line and it kept playing over the caller.

Observed on live calls: after two or three consults the assistant "stumbles" and turns into mumbling, then answers get cut mid-sentence.

## Fix

- `OpenAIRealtimeClient` tracks every open response by id and cancels each one explicitly on barge-in. `hasActiveResponse` is now correct with concurrent responses.
- `create_response` is switched off while a consult is in flight and back on right before the tool output is posted. The caller's words still land in the conversation and are answered together with the tool result.
- `onReady` fires only for the first `session.updated`, so the toggle round-trips do not replay the greeting.
- The reassurance instruction tells the model it has no answer yet and must not guess.
- The client accepts optional VAD tuning (`threshold`, `prefixPaddingMs`, `silenceDurationMs`); the phone surface will use it in a follow-up PR.

## Tests

Four new cases in `tests/voice.realtime-bridge.test.ts`: cancel of all in-flight responses by id, per-id active accounting, auto-response suspend/resume ordering around the tool output without re-greeting, and the no-guessing reassurance text. Existing voice suites pass; the two unit failures on `main` (eval-command, host-runner) are unrelated.

### DIFF
--- a/src/channels/voice/openai-realtime.ts
+++ b/src/channels/voice/openai-realtime.ts
@@ -67,6 +67,21 @@ function normalizeString(value: unknown): string {
   return typeof value === 'string' ? value : '';
 }
 
+// Responses the server opened without an id (older event shapes, tests) are
+// tracked under a synthetic key so cancel/accounting still balance.
+const ANONYMOUS_RESPONSE_PREFIX = 'anon:';
+
+function responseId(event: Record<string, unknown>): string {
+  return isRecord(event.response) ? normalizeString(event.response.id) : '';
+}
+
+/** Server-side VAD tuning; omitted fields keep the upstream defaults. */
+export interface RealtimeTurnDetection {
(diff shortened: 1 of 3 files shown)


maxnoller:
Folded into #1503, which now targets main and carries both changes.

PR #1507, opened 2026-09-10
fix(memory): persist tool calls and results across turns

furukama:
## Summary

- Follow-up turns lost the tool calls and results behind the assistant's earlier reply. Persist ordered exchanges with each assistant message and replay them when rebuilding context, including in a fresh worker.
- Bound individual results in model context at 16,000 characters, preserve context-guard pruning, and retain full credential-redacted results in searchable session transcripts. `session_search` returns transcript paths and supports searching by call ID.
- Preserve call/result pairs through approvals, error turns, forks, token budgeting, and compaction. Error results use the existing error-turn persistence path to avoid duplicate messages.
- Approval policy, tool permissions, audit authority, and canonical cross-channel memory are unchanged. This wires gateway chat and bootstrap turns; scheduler/heartbeat persistence is outside this change.

## Change Type

- [x] Bug fix
- [x] Docs
- [x] Tests

## Linked Context

Fixes #1469

## Licensing

- [x] All commits are signed off (`git commit -s`) to certify the [DCO](https://developercertificate.org/); the contribution is licensed under the repository's MIT license.

## Validation

- `npm run format`, `npm run lint`, `npm run typecheck`, and `npm --prefix container run lint` pass. Formatting reports existing optional-chain warnings outside this change.
- Console, gateway, and container builds pass. The console uses `vite build --configLoader runner`, followed by `npm run build --ignore-scripts`, because shared read-only dependencies prevent Vite from writing its temporary config bundle.
- All 124 targeted tests across eight suites pass: tool history, database sessions, gateway media history, memory recall and citations, memory service behavior, and compaction routing and summaries. This covers replay alongside skipped empty-memory recall and summary-only access.
- The schema-58 upgrade regression verifies that tool history is added while existing chat, scheduler errors, delivery failures, and the scheduler migration record remain intact; repeated initialization is idempotent.
- A worker-process integration test reads a file, persists the result, reopens SQLite, deletes the original file, and starts a fresh worker; the next request receives the stored result without repeating the tool call.
- Boundary tests cover malformed roles and provider metadata, incomplete exchanges, result caps, redaction, transcript symlinks, session separation, forks, and failed turns without duplicate history entries.
- Live-provider and Docker runtime checks were not run. Provider adapters had targeted unit coverage; the IPC integration runs the actual worker in separate Node processes.

## Docs And Config Impact

- [x] Developer memory documentation updated.
[...]

PR #1512, opened 2026-09-16
Give agents their real workspace path instead of a /workspace alias

maxnoller:
## Problem

In host mode the agent runtime was told its workspace lived at `/workspace` while the files were somewhere else on disk (`<data>/agents/<id>/workspace`). File tools translated that alias, `bash` did not, and the system prompt used the real path — so the model saw three different truths:

- `ls /workspace` showed nothing familiar (it is the *parent* of the data dir in the cloud image), so the agent started hunting.
- Uploaded files were given as `/uploaded-media-cache/…`, which does not exist for shell commands or the skill scripts (`skills/pdf/scripts/*.mjs …`) → "agent can't find the uploaded file".
- Outputs written to `/workspace/out.pdf` from bash landed outside the real workspace, where artifact discovery never walks → "PDF was created but nothing attached".
- Because the real data dir started with the alias prefix, real paths could be mis-remapped; several guards existed only to compensate for that.

## Change

**Real paths everywhere.** The agent runtime addresses its workspace and the media caches at their real location. `HYBRIDCLAW_AGENT_WORKSPACE_DISPLAY_ROOT` and `workspaceDisplayRootOverride` are removed along with the collision guards and the "use relative paths instead of literal /workspace paths" hedge in the bash tool description.

The only remaining mapping is the task-sandbox eval mode, where bash is proxied into a container that mounts the workspace elsewhere; that display root is now derived from `HYBRIDCLAW_BASH_DOCKER_CWD` instead of being a separate knob.

Gateway-side resolvers (Discord `send_file`, PDF context, artifact remap) prefer real paths and keep the `/workspace` mapping only for the container executor, where it is a genuine bind mount. Host mode no longer applies it.

**Short, intuitive real path.** New optional `HYBRIDCLAW_WORKSPACES_DIR` places workspaces at `<dir>/<agentId>`. The cloud image sets it to `/workspace`, so the default agent works in `/workspace/main` (and `ls /workspace` shows the agents). Existing workspaces are moved there once — at gateway startup for registered agents and lazily on first use — with a no-op when the target already exists.

Net: −208/+169 lines before tests.

## Verification

- `tsc` (root + container), `tsc --noUnusedLocals`, biome: clean on changed files.
- Full unit suite: 5802/5803 pass; the one failure (`host-runner.redaction` → agent-browser binary path) fails identically on `main` in this checkout.
- New test: `agentWorkspaceDir` honours `HYBRIDCLAW_WORKSPACES_DIR` and moves the legacy workspace exactly once.
- Tests that encoded the alias now assert real paths.

Not yet verified on a live sandbox — that is the next step after review.

## Follow-ups (outside this repo / PR)

- Anything provisioning a sandbox that writes into `<data>/agents/<id>/workspace` directly (e.g. seeding `USER.md`) must switch to `/workspace/<id>` once this rolls out; until then such writes are moved along with the workspace on the next boot only if the new dir does not exist yet.
- Artifact links in existing chat history point at the old workspace path and will 404 after the move.
- A read-only file browser in the console becomes trivial now that there is only one path.


maxnoller:
@claude please review this pr

maxnoller:
Addressed the review in 9bd45edd:

**Silent data loss on a failed migration** — fixed. `migrateLegacyAgentWorkspace` now throws instead of logging and returning, so `ensureAgentDirs` / `ensureBootstrapFiles` never create an empty target after a failed move and the next boot retries. `EXDEV` (workspaces dir on another filesystem) falls back to a staged copy: copy into a `<target>.migrating` sibling, rename into place, then remove the legacy dir, so a crash mid-copy never leaves a partial target the existence guard would accept. Two new tests in `tests/ipc.test.ts` cover the EXDEV fallback and the failure path (throws, no empty target, legacy intact, later boot migrates). Note the gateway startup loop also propagates the error, so a broken `HYBRIDCLAW_WORKSPACES_DIR` stops boot rather than starting agents fresh.

**Tool error strings now show real paths** — intended; the model needs the path that actually resolves.

**Live verification** — the problem itself is now confirmed from prod traces (Langfuse, 2026-09-14): the model called `read /uploaded-media-cache/<file>` from the tool description → `Path escapes workspace`, then `cd /workspace` (a real dir in the cloud image), wrote its output there, `read /workspace/X.docx` → `File not found`, and the file was never attached. That session already ran with #1507's tool-history, which did not prevent it. Live smoke on staging after rollout remains a follow-up.

PR #1520, opened 2026-09-18
fix(msteams): retry inbound attachment downloads and tell the model when one never arrived

maxnoller:
## Problem

When a Teams inline image/file could not be staged into the media cache, `buildMediaItem` caught the error, logged it at **debug** level and returned the URL-only fallback item (`path: null`, `size: 0`, `mime: unknown`). Tools cannot fetch Bot Framework attachment URLs, so the model was left with a question whose subject was in the image and no way to see it — and it answered anyway. We saw exactly this in production: the same short German question sent twice with a screenshot each time, both downloads failed silently, and the second reply confidently described an unrelated Excel export from earlier in the session.

## Fix

- `buildMediaItem` now retries the remote download (3 attempts, 500 ms / 1.5 s backoff). A 401/403 evicts the cached client-credentials token so the retry mints a fresh one instead of replaying a stale token. Oversized attachments still bail out immediately.
- The final failure is logged at **warn** with attempt count and reason.
- The media item carries `unavailableReason`, and `buildMediaPromptContext` renders an `UnavailableMedia` block instructing the model that the attachment did not arrive and to ask the user to resend it — instead of guessing.

## Tests

- `tests/msteams.attachments.test.ts` (new): transient failure → retry → staged; exhausted retries → `unavailableReason` + warn log; oversized → no retry.
- `tests/gateway-service.media-prompt-context.test.ts`: `UnavailableMedia` block present/absent.
- `tsc --noEmit --noUnusedLocals --noUnusedParameters` and the container typecheck pass; `vitest run tests/msteams tests/gateway-service tests/gateway-main.test.ts` → 463 passed.


maxnoller:
@claude please review this pr

maxnoller:
Addressed in the follow-up commit:

1. **Double fetch / latency amplification** — the reported size is now checked *before* any download; an oversized attachment is never fetched (or retried). Each branch's result (staged or `unavailableReason`) is final for that attachment, so a failed `downloadUrl` no longer falls through to `contentUrl`. Covered by two new tests (`skips the download when Teams reports a size over the limit` → zero fetches; `does not fetch contentUrl after the download info fetch already failed` → exactly 3 calls, all to the download URL).
2. **Cache eviction** — kept `clear()` (the failing scope isn't known at that point and there are only two); added a comment saying so.
3. **Header** — updated to describe the retry and the `unavailableReason` contract consumed by `gateway-service.ts`.
4. **Oversized asymmetry** — oversized attachments now come back as unavailable items with reason `larger than the N MB media limit`, so the model can tell the user why the file didn't arrive. The one exception is an oversized inline *data URL* (personal-chat inline image above the limit): that still returns nothing, because the fallback item would carry the multi‑MB data URL into the prompt's `MediaUrls`.

PR #1524, opened 2026-09-18
fix: align A2A JSON-RPC ids and error codes with the A2A spec

furukama:
## Summary

- Problem: A check against the [A2A v1.0 Send Message spec](https://a2a-protocol.org/latest/specification/#311-send-message) found three places where HybridClaw's A2A JSON-RPC adapter breaks basic protocol conventions, even between HybridClaw peers:
  - Outbound `message/send` requests had no JSON-RPC `id`, which makes them notifications. A compliant server may send no response.
  - Outbound requests set `message.taskId` to `parent_message_id`. A2A requires a client-supplied `taskId` to reference an existing task that the server issued, so a standard server returns `TaskNotFoundError` for every threaded reply.
  - Inbound auth failures returned `-32001` and duplicates returned `-32009`. A2A reserves those codes for `TaskNotFoundError` and `VersionNotSupportedError`.
- Why it matters: These are the low-cost fixes that stop HybridClaw from sending misleading protocol signals. There's also a real bug: a duplicate message got HTTP 409. When a first response was lost and the sender retried, the sender marked an already-delivered message as failed.
- What changed:
  - Every outbound request carries `id: <envelope id>`.
  - `taskId` is no longer sent. Reply threading stays in the HybridClaw envelope metadata, which receivers already read.
  - Auth failures return `-32000`, the generic JSON-RPC server error, because A2A defines no auth code.
  - A replayed message returns the stored delivery receipt with HTTP 200 and `already_delivered: true`. This matches `/a2a/envelopes` and the spec's `messageId` idempotency guidance.
- What did not change: This PR does not make HybridClaw a conforming A2A Send Message server or client. Still out of scope:
  - The inbound side still requires the HybridClaw envelope in `metadata`.
  - Responses are a delivery receipt, not a `Task` or `Message`.
  - There is no task model, no `A2A-Version` handling, and no v1.0 method names or agent card.
  - The webhook transport's 409 on duplicates is untouched.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [x] Tests
- [ ] Refactor required for the fix
- [ ] Tooling or workflow
- [ ] Security hardening

## Linked Context

- Closes #
- Related #

## Licensing

- [x] All commits are signed off (`git commit -s`) to certify the
      [DCO](https://developercertificate.org/); the contribution is licensed
      under the repository's MIT license.

## Validation

```bash
npm run format   # no fixes applied
npm run lint     # clean
npx vitest run --configLoader runner --project unit tests/a2a-          # 12 files, 118 tests passed
npx vitest run --configLoader runner --project integration tests/a2a-outbound.integration.test.ts  # 4 passed
```

- Verified manually: n/a. This is protocol-level behavior and the tests cover it.
- Edge cases checked:
  - The replay receipt matches the first delivery's canonical `recipient_agent_id`, not the local alias from the request.
  - Error responses now echo the request `id` instead of `null`.
- Skipped checks and why: I didn't run the full `npm run test:unit`. The change is limited to `src/a2a/a2a-inbound.ts` and `src/a2a/a2a-json-rpc.ts`, and all `tests/a2a-*` suites pass.

## Docs And Config Impact

- [ ] README, docs, or examples updated
- [ ] Config or environment behavior changed
- [ ] Templates or workspace bootstrap files changed
- [x] No docs or config impact

CHANGELOG `Unreleased → Fixed` entry added.

## Risk Notes

- Security-sensitive paths touched? Yes. `src/a2a/a2a-inbound.ts` handles inbound peer requests. Authentication and authorization logic is unchanged; only the JSON-RPC error code for auth failures changed (`-32001` → `-32000`). The HTTP status is still 401.
- Gateway, audit, approval, or container boundaries touched? Audit only. The inbound audit event for a duplicate now records `statusCode: 200` instead of 409. `downstreamDisposition: 'duplicate'` is unchanged.
- Mixed-version peers:
  - Older senders that omit `id` are still accepted.
  - Older senders now get a 200 receipt for a duplicate, so they mark it delivered instead of failed.
  - For auth failures, older senders fail fast on the HTTP 401 before reading the JSON-RPC code.

## Evidence

- [ ] Failing output before and passing output after
- [ ] Screenshot or recording
- [ ] Log snippet
- [x] New or updated test coverage:
  - New inbound test: `answers a replayed JSON-RPC message idempotently`
  - New outbound test: `keeps reply threading in the envelope instead of the A2A taskId`
  - Updated tests: auth-error code and echoed request `id`

## Follow-up

`shouldRetryA2AJsonRpcErrorCode` in `src/a2a/a2a-retry-policy.ts` retries everything in `-32000..-32099`. That range includes A2A's permanent errors (`-32001`, `-32004`, `-32005`, `-32009`). Against a non-HybridClaw A2A server, delivery would retry those until it hits max attempts. That's worth a separate small fix.


furukama:

PR #1526, opened 2026-09-20
feat(discord): add configurable threaded replies

### LINKED ISSUE #1514: feat(discord): configurable reply style — answer in a thread instead of the channel
## Problem

There is no way to configure the Discord channel to answer **in a thread** instead of posting into the channel. Teams and Slack both have this (`msteams.replyStyle`, `slack.replyStyle`, `'thread' | 'top-level'`, defaulting to `thread`), Discord has nothing equivalent — so in a busy guild channel every agent turn lands inline and pushes the human conversation up.

## Current behaviour

- `RuntimeConfig['discord']` and `RuntimeDiscordChannelConfig` (`src/config/runtime-config.ts`) have no `replyStyle`; there is no `DiscordReplyStyle` type next to `MSTeamsReplyStyle` / `SlackReplyStyle`, and nothing in the console settings registry.
- The reply goes wherever the trigger came from, because the send target is `msg.channelId`. So threads work *if a human starts one and writes inside it* — inbound already pulls the `[Thread starter]` message into the prompt (`src/channels/discord/runtime.ts:1005`) and delivery checks `SendMessagesInThreads` (`src/channels/discord/runtime.ts:2759`) — but that is user-side behaviour, not configuration.
- The agent can create/reply in threads through the discord tool actions `thread-create` / `thread-reply` (`src/channels/discord/tool-actions.ts:1399`, `skills/discord/SKILL.md`), so it can be prompted into it via channel instructions. That does not solve the problem: the turn's own streamed text still goes to the origin channel and the thread post is an extra message on top.

## Proposal

Mirror the Teams/Slack shape:

1. `export type DiscordReplyStyle = 'thread' | 'top-level'` in `src/config/runtime-config.ts`.
2. `discord.replyStyle` (global, default `'top-level'` to keep current behaviour) plus a per-channel override in `RuntimeDiscordChannelConfig`, resolved in `resolveChannelBehavior` (`src/channels/discord/runtime.ts:757`) the way the other per-channel knobs already are.
3. `DISCORD_REPLY_STYLE` in `src/config/config.ts` + normalizer in the config parser, so config.json and env both work.
4. Delivery: when the style is `thread`, the message is in a guild channel and it is not already a thread, start a thread off the triggering message (`message.startThread`, as `runDiscordThreadCreateAction` already does) and stream the turn into it; reuse the thread on follow-ups. DMs and messages already inside a thread keep today's path.
5. Needs `CreatePublicThreads` (and `SendMessagesInThreads`) — the existing permission preflight should cover it and degrade to a channel reply with a warning rather than dropping the answer.
6. Thread name: derive from the triggering message, since Discord requires a name at creation.

## Caveat worth fixing alongside

Channel mode is resolved by exact `msg.channelId` with no parent fallback (`src/channels/discord/runtime.ts:747`). A thread has its own id, so with `groupPolicy: 'allowlist'` a freshly created thread is not allowlisted and the agent goes silent inside it — including a thread it just opened itself. Thread ids should fall back to the parent channel's config for mode and send-policy lookups.

maxnoller:
## Summary

- add global and per-channel Discord replyStyle configuration, with DISCORD_REPLY_STYLE environment override and console controls
- lazily create or reuse a public thread for guild conversation output while preserving DMs and messages already in threads
- inherit inbound mode and outbound send policy from a thread's parent channel, with exact thread overrides still taking precedence
- fall back to the source channel with a warning when thread creation or permissions are unavailable

## Risk and failure modes

- default remains top-level, so existing installations keep current behavior
- thread creation requires CreatePublicThreads and SendMessagesInThreads; missing permissions are preflighted and degrade to the existing channel reply path
- thread creation is delayed until the first output chunk, avoiding empty threads for silent turns

## Validation

- npm run lint
- npm run typecheck
- npm run check (passes with existing repository warnings)
- npx vitest run tests/discord-stream.test.ts tests/discord.thread-replies.test.ts tests/discord.send-permissions.test.ts tests/configured-models.test.ts
- npx vitest run tests/discord.tool-actions-send.test.ts tests/discord.tool-actions-channel-lookup.test.ts
- npm --prefix console test -- --run src/routes/channels.test.tsx

Closes #1514


maxnoller:

Issue #1468, opened 2026-09-03
Chat history is cut at 40 messages / 24k chars without a marker or summary before compaction


maxnoller:
## Problem

Each turn sends at most 40 stored messages (`MAX_HISTORY_MESSAGES`, `src/gateway/gateway-chat-service.ts:201, 1662-1665`), further cut to 24,000 characters by dropping whole oldest turns (`src/session/token-efficiency.ts:4, 190-238`). Nothing in the prompt indicates that earlier turns were removed; the drop counts only go to an audit event. `HEAD_TAIL_TRUNCATED_MARKER` is used for per-message truncation only.

No summary exists until session compaction fires, which happens at 200 stored messages or roughly 70k estimated tokens (`src/config/config.ts:709-711`, `session-maintenance.ts:337-390`). Between message 41 and compaction, older context is simply absent. The only bridge is "Relevant Memory Recall": at most 5 snippets of 220 characters (`src/memory/memory-service.ts:286, 368`) drawn from previous compaction summaries and audio transcripts, so it is empty for a session that has never been compacted. Those `[mem:N]` citations are chat recall, not files, but the model presents them as evidence of "memory".

## Proposed fix

- Insert an explicit "[earlier N turns omitted]" marker when turns are dropped.
- Run a lightweight rolling summary once the 40-message window is exceeded, or align the compaction threshold with the prompt window.
- Label recall snippets as "chat recall" in the prompt so the model does not cite them as saved memory.

---
Source: HybridClaw Reliability Assessment (2026-09-03, read-only audit against v0.29.1), re-verified against main v0.30.0.

maxnoller:
Implementation split into two PRs:

- #1531 — one token budget shared by prompt history and compaction, turn-aligned retention, and a `## History Window` note when turns are still dropped (points 1–2, plus the marker).
- #1532 — recall block relabeled `### Chat Recall` and per-turn snippets skipped while their turn is still verbatim history (point 3).

maxnoller:
Third PR added from the design review: #1533 runs automatic compaction and `/compact` through one engine (stacked on #1531). #1532 also now skips compaction-summary rows from recall while the session summary is injected.

maxnoller:
Update: #1533 was folded into #1531, so the work is two PRs: #1531 (shared budget, history window note, one compaction engine) and #1532 (chat recall labeling and de-duplication).

PR #1533, opened 2026-09-21
refactor: run automatic and manual compaction through one engine

maxnoller:
## Summary

Stacked on #1531. Follow-up to the design review on #1468: there were two compaction engines with different retention rules.

- `src/session/session-maintenance.ts` (automatic) built its own flat summary prompt, deleted rows by cutoff id, and wrote a JSONL export.
- `src/memory/compaction.ts` (manual `/compact`) produced the structured summary, kept three recent messages by ratio, archived the transcript, and stored a compaction memory.

Now there is one path:

- `compactSessionNow` in `session-maintenance.ts` runs plugin hooks and the pre-compaction memory flush, then calls `memoryService.compactSession` with the turn-aligned, token-bounded retention count from #1531. `maybeCompactSession` is only the trigger check on top of it.
- `/compact` in the gateway calls `compactSessionNow` instead of the engine directly, so manual and automatic compaction keep the same slice, produce the same structured summary, archive the transcript, and fire the same plugin hooks.
- `compactConversation` accepts `retainRecentCount`; when set it replaces the ratio-based split.
- The legacy summary prompt and `exportCompactedSessionJsonl` are removed. The transcript archive under the data directory is the record of compacted history.

## Risk notes

- `src/gateway/gateway-service.ts` changes only the `/compact` command handler.
- The compaction summary is now the structured six-section markdown for automatic runs as well. Anything parsing `session_summary` as free text keeps working; it is still text.

## Test plan

- [x] `npm run lint`
- [x] `tests/session-maintenance.test.ts` rewritten around the shared engine, including `compactSessionNow` returning null when nothing is compactable
- [x] `tests/memory-compaction.test.ts` covers explicit retention
- [x] heartbeat, plugins, http-server, doctor, and compaction-router suites pass

Refs #1468


maxnoller:
Folded into #1531.

PR #1540, opened 2026-09-21
feat: unified model routing with JEV shadow comparisons

furukama:
## Unified model routing

Models → Routing configures shared capability tiers, mode-specific model assignments, an optional live classifier and an optional comparison classifier. Both classifiers select a tier; the execution policy applies privacy boundaries, explicit pins, escalation and safe fallback. Chat exposes router decisions, model, timing, tokens, cache usage and costs.

Privacy boundaries are Local, HybridAI, EU provider, EU hosting and World. Save validation rejects unusable routes. Cost uses known token rates; Speed and Auto use recent successful tool-free execution timings. Incomplete timing coverage preserves configured order rather than dropping unmeasured candidates.

Classifiers are opt-in. Upgrades preserve a disabled v1 concierge, and adding a JEV key alone does not enable comparison. Labs sample approval remains separate from configured live comparison. Classification has its own auxiliary task, independent of Skills Hub. The obsolete concierge metadata and Discord profile-button paths are removed.

Shadow classification runs alongside agent execution and cannot change its model or fail the live turn. Final accounting waits for bounded shadow completion to retain comparison usage. Classifier disclosure guards skip classification without silently changing the saved execution boundary.

## Validation and boundaries

- Lint/type checks passed for gateway, console and desktop.
- 166 targeted backend tests and 30 console tests passed, including upgrade opt-out, comparison consent, privacy denial before transport, invalid tiers, low confidence, pending/failed shadow calls, timing coverage, and tool-free latency sampling.
- Privacy limits still constrain live, shadow, pinned and fallback model calls. Disclosure heuristics are not PII guarantees; execution follows the operator's saved privacy boundary.
- Latency remains a process-local estimate. No live-provider benchmark or full live-provider suite was run for these review fixes.


maxnoller (commented):
## Review

### 🐛 Bugs

**1. Routed turns never record latency** · `gateway-chat-service.ts:2585`
`recordRoutingLatency` only runs in the `!routingAttempts` branch. With routing on, every unpinned turn goes through `executeModelRouting`, so speed/auto only ever see timings from *pinned* turns (which also include tool time).

**2. Upgrade silently enables the live classifier** · `runtime-config.ts:7117`
Existing configs have `concierge: { enabled: false, model: "gemini-3-flash" }` persisted (shipped since v0.10.0). The new normalizer keeps `model` and drops `enabled` → routing-enabled installs start a gemini-3-flash classifier call on every interactive turn. Needs a migration (clear `model` when `enabled` was false).

**3. Classifier depends on `skills_hub` aux settings** · `unified-routing.ts:124`
`task: 'skills_hub'` → if an admin disables that task, `resolveTextCallContext` throws and every classification silently becomes `classifier-failed`. Deserves its own task (or no task gate).

**4. Concierge v1 leftovers** · `discord/runtime.ts:262,1615` · `gateway-chat-service.ts:678,694,1455`
The plugin no longer emits `conciergeRouter` metadata, so that branch is dead. Old Discord concierge buttons still work and now send `asap` / `balanced` / `no_hurry` to the agent as a plain user message. Per CLAUDE.md, remove rather than carry forward.

### ❓ Intended?

**5. Local-only trigger is very broad** · `unified-routing.ts:42`, `routing/evaluator.ts:21`
Any disclosure reason forces local execution. The regex matches `2026-09-23`, `1000 + 2000 + 3000`, `make this method private`, any email, `ignore the lint rules`; >4000 chars also counts. On cloud-only setups these fail with *"No eligible local model"*, and a pinned cloud model fails as soon as an image is attached. Is that acceptable, or should only real sensitivity hits force local (and attachments/length just skip the classifier)?

**6. One measured model wins speed/auto** · `routing/policy.ts:118`
Unmeasured candidates are dropped entirely, so a single timed model (e.g. opus from one pinned turn) becomes the start model for everything. Combined with #1 this is likely in practice.

**7. Shadow classifier blocks the live turn** · `gateway-chat-service.ts:1594`
`Promise.all` waits for the comparison call (up to its timeout). If awaiting is only needed to show both tags, consider recording the shadow result async. Also: secret store is read + decrypted 2–3× per turn (`isJevAvailable` + `evaluateConfiguredRouting`).

**8. JEV receives live prompts once a key exists** · `gateway-chat-service.ts:1604`
`comparisonModel` defaults to `jev/jev-latest` and the shadow call hard-codes `publicSample: true`, bypassing `evaluator.mode: 'off'`. Documented in `typed-routing-evaluator.md`, but conflicts with `model-routing.md` §8 ("opt-in … exact admin-approved public prompts"). Should adding the key really be the opt-in?

furukama:
@maxnoller Check again, added some fixes

PR #1544, opened 2026-09-22
feat(auth): sign in to HybridAI with OAuth 2.1 instead of a pasted API key

maxnoller:
## Summary

Onboarding and `hybridclaw auth login hybridai` now sign the user in with OAuth 2.1 against the HybridAI platform instead of asking them to paste a long-lived platform API key:

1. RFC 8414 discovery on the configured base URL (`https://hybridai.one` by default)
2. RFC 7591 registration of a public client with a loopback redirect (`http://127.0.0.1:<random port>/oauth/callback`)
3. Authorization code + S256 PKCE, consent in the browser, callback on the loopback listener
4. Scoped, revocable access token (`profile api mcp`) plus a rotating refresh token

Headless shells (SSH sessions, containers) use the OAuth 2.0 device authorization grant (RFC 8628) instead, since the loopback redirect never reaches them:

```
HybridAI sign-in
On any device with a browser, open:
  https://hybridai.one/device
and enter the code:  WXKT-QMBD
(or open https://hybridai.one/device?user_code=WXKT-QMBD directly)
Waiting for approval (code valid for 15 min) ...
Signed in as max@example.com.
```

`hybridclaw auth login hybridai --device-code` (the default on such shells) and onboarding there use it. If the platform's discovery metadata has no `device_authorization_endpoint`, they fall back to the loopback flow with the sign-in URL printed and the redirect URL (or bare code) pasted back into the terminal; whichever arrives first wins.

## Design

- **Every existing consumer keeps working unchanged.** The access token is stored as `HYBRIDAI_API_KEY` in the encrypted runtime secret store; the platform accepts these tokens wherever it accepts an API key. Refresh material and client metadata live next to it as `HYBRIDAI_OAUTH`.
- **Token rotation.** The gateway runs `startHybridAIAccessTokenMaintenance` (one-minute tick, refresh when less than 15 minutes remain, single in-flight refresh). Every CLI start (`ensureRuntimeCredentials`) refreshes a stale token first, so `tui`/`gateway` after a long pause come up with a valid one. `invalid_grant` (revoked/expired/reused refresh token) removes the stale access token so callers get a clear "not configured" error instead of opaque 401s; transient failures keep everything and retry.
- **Pasted keys still win.** `--api-key` keeps the previous paste flow. Storing a platform key over an OAuth session (CLI, console, `secret set`) drops the session; the key is authoritative.
- **Fail-fast fallback.** If the base URL publishes no OAuth metadata (a platform without the authorization server), onboarding says so and continues with the API-key flow; `auth login hybridai` points at `--api-key`.
- **Logout revokes** the refresh token family at the platform (best-effort) before clearing the store.
- `src/auth/oauth2-client.ts` holds the OAuth primitives (discovery, registration, PKCE, token, revocation, device authorization + polling) shared with the remote MCP server client, which is refactored onto it with unchanged behaviour.
- **Device flow.** `startHybridAIDeviceAuthorization` registers a public client with the `refresh_token` + device grant types (registration requires a redirect URI, so a loopback placeholder is sent; the device flow never uses it), then `pollDeviceToken` handles `authorization_pending`, widens the interval on `slow_down`, rejects on `access_denied` / `expired_token`, retries transport errors until the code expires, and is abortable. The stored session is identical to the code flow's, so refresh, status and logout do not know which grant produced it.

## Threat / risk notes (touches `src/gateway/`, secrets)

- Tokens never hit logs or stdout; only the secret store (encrypted) holds them. Status output masks them like API keys.
- Loopback listener binds 127.0.0.1 on an ephemeral port, validates `state`, answers a single request, and dies with the flow (10-minute timeout). A foreign `state` is ignored, not fatal.
- PKCE binds the code to this process; a pasted foreign code fails at the token endpoint.
- Redirect URIs are loopback `http://` only, as the platform's registration policy requires.

## Verification

- New `tests/hybridai-oauth.test.ts`: registration + PKCE exchange through a real loopback callback, pasted-redirect and bare-code paths, state mismatch, fail-fast without metadata/registration, refresh (fresh / rotate / `invalid_grant` sign-out / transient keep), session dropped by a pasted key, revoke-on-logout; device flow: request/registration shape, poll sequence pending → slow_down → tokens (fake timers), single start, denial, expiry, cancel, null on platforms without the endpoint.
- `tests/onboarding.test.ts`: full interactive OAuth onboarding (stubbed platform), headless onboarding via device code, plus the existing API-key onboarding now exercised as the fallback.
- Updated `tests/hybridai-auth.test.ts` (incl. `--device-code` through the device flow), `tests/cli.test.ts`; `tests/mcp-oauth.test.ts` unchanged and green after the refactor.
- `npm run lint`, `npm run check`: clean. Full unit suite green except `host-runner.redaction` "exposes the bundled agent-browser binary", which fails identically on `main`.
- Live: discovery, registration, consent redirect through login and the loopback callback verified against a platform build that ships the authorization server; the production platform does not publish OAuth metadata yet, so onboarding there falls back to the API-key flow (verified).

## Follow-ups

- Access tokens live one hour and are rotated at 15 minutes left; a single agent run longer than that could hit an expired token mid-run. A longer access-token lifetime on the platform side would remove that edge.
- Each login registers a new client (the loopback port is random). Fine for interactive use; worth a fixed port or a reusable client if logins become frequent.


maxnoller:
### Sign-in UX demo

Recorded against a local stub authorization server (`127.0.0.1:5055`) with a throwaway `demo@example.com` account. Tokens are fake and the paths are anonymized.

1. **Desktop:** `auth login hybridai` opens the consent page. The browser redirect lands on the loopback listener and the token is validated.
2. **`auth status`:** method `oauth`, the account, and when the access token expires (it refreshes automatically).
3. **Headless:** `--device-code` prints a user code for `/device` and polls until the user approves.
4. **Logout:** revokes the OAuth session at the platform, then clears the local credentials.

HybridAI OAuth sign-in demo

PR #1545, opened 2026-09-22
feat(auth): device code sign-in for headless shells (RFC 8628)

maxnoller:
## Summary

Stacked on #1544. On SSH sessions and in containers the loopback redirect from the OAuth code flow never reaches the gateway host, so the user had to copy a "connection refused" URL out of the browser's address bar. This adds the OAuth 2.0 device authorization grant (RFC 8628) for those cases:

```
HybridAI sign-in
On any device with a browser, open:
  https://hybridai.one/device
and enter the code:  WXKT-QMBD
(or open https://hybridai.one/device?user_code=WXKT-QMBD directly)
Waiting for approval (code valid for 15 min) ...
Signed in as max@example.com.
```

`hybridclaw auth login hybridai --device-code` (the default on headless shells) and onboarding on such shells use it. Desktops keep the loopback flow with the auto-opened consent page. Platforms whose discovery metadata has no `device_authorization_endpoint` fall back to the previous loopback flow with a pasted redirect, so this is safe to ship ahead of the platform.

## Design

- `src/auth/oauth2-client.ts`: `requestDeviceAuthorization` and `pollDeviceToken` (handles `authorization_pending`, widens the interval on `slow_down`, rejects on `access_denied` / `expired_token`, retries transport errors until the code expires, abortable).
- `src/auth/hybridai-oauth.ts`: `startHybridAIDeviceAuthorization` registers a public client with the `refresh_token` + device grant types (registration requires a redirect URI, so a loopback placeholder is sent; the device flow never uses it) and stores the resulting session exactly like the code flow (`HYBRIDAI_API_KEY` + `HYBRIDAI_OAUTH`), so refresh, status and logout are unchanged.
- The stored session is identical to the code flow's; nothing downstream knows which grant produced it.

## Verification

- `tests/hybridai-oauth.test.ts`: device authorization request and registration shape, poll sequence pending → slow_down → tokens (fake timers), session stored, single start, denial, expiry, cancel, and null on platforms without the endpoint.
- `tests/hybridai-auth.test.ts`: `--device-code` login through the device flow, printed instructions, resulting status.
- `tests/onboarding.test.ts`: onboarding on an SSH shell signs in with a device code and never prompts for a browser.
- `npm run lint`, `npm run check` clean; full unit suite run before opening this PR.


maxnoller:
Folded into #1544 (same commits, fast-forwarded onto that branch).

